PR #6, opened 2025-09-14
fix login endpoint by changing it from get to post

AimanYosofiUlt:


nabilBouzineDev (approved):

ShroukMohamed16 (approved):

identity/src/main/kotlin/identity/controller/IdentityController.kt
```diff
@@ -15,8 +18,12 @@ import org.springframework.web.bind.annotation.RestController
 class IdentityController(
     val authenticationService: AuthenticationService
 ) {
+    @ExceptionHandler(BadCredentialsException::class)
+    fun handleBadCredentialsException(): ResponseEntity<String> {
+        return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body("Invalid phone number or password")
+    }
```
Karrar-Mohammed: why do we have the exception handler here in the controller, i suggest to create a controller advice for handling the exceptions and put this function inside the exception handler

mohamedd-hassan (approved):

identity/src/main/kotlin/identity/config/exceptionHandling/GlobalExceptionHandler.kt
Karrar-Mohammed, line 1: i didn't ask you to do all the error handling, i aske to only do it for your team module, and you just copy pasted it from the evolve fit that we don't need it exactly the same

Karrar-Mohammed:
the expected from this PR i just one line of change from @GetMapping to @PostMapping

identity/src/main/kotlin/identity/config/exceptionHandling/GlobalExceptionHandler.kt
AimanYosofiUlt: Ok, I will change it to `IdentityExceptionHandler`, which contains only login Exceptions

AimanYosofiUlt: Ok, I’ll rename it to IdentityExceptionHandler, which **now** handles only login-related exceptions.

AimanYosofiUlt:
> the expected from this PR i just one line of change from @GetMapping to @PostMapping

I’ve removed all changes unrelated to this PR—sorry about that!

Karrar-Mohammed (approved):

PR #7, opened 2025-09-15
Feature/WLT-10 create tables and repository

Malak187:
- Implemented entity models for Wallet, Transaction, Block, and TransactionBlock to represent the blockchain structure in the database.
- Designed the **TransactionRepository** with a query method to retrieve all transactions for a given user by checking their ID in both sender and receiver fields.


wallet/src/main/kotlin/wallet/entity/Transaction.kt
```diff
@@ -0,0 +1,33 @@
...
+    val id: UUID = UUID.randomUUID(),
+
+    @Column(nullable = false, updatable = false)
+    val timestamp: LocalDateTime = LocalDateTime.now(),
+
+    @Column(name = "sender_id", columnDefinition = "uuid", nullable = false, updatable = false)
+    val senderId: UUID,
+
+    @Column(name = "receiver_id", columnDefinition = "uuid", nullable = false, updatable = false)
+    val receiverId: UUID,
```
Al-Qassim: using wallet address for sender and receiver, instead of user id, will be better, since this is what the wallets are for

wallet/src/main/kotlin/wallet/entity/TransactionBlock.kt
```diff
@@ -0,0 +1,23 @@
+package wallet.entity
+
+import jakarta.persistence.*
+import java.io.Serializable
+import java.util.*
+
+@Entity
+@Table(name = "transaction_block")
+@IdClass(TransactionBlockId::class)
+data class TransactionBlock(
```
Al-Qassim: naming it `transaction_block_cross_ref` `TransactionBlockCrossRef` with make it clearer

wallet/src/main/kotlin/wallet/entity/Transaction.kt
Malak187: makes sense. I'll edit it.

Al-Qassim (approved):

wallet/src/main/kotlin/wallet/entity/Wallet.kt
```diff
@@ -0,0 +1,18 @@
...
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.util.*
+
+@Entity
+@Table(name = "wallets")
+data class Wallet(
+    @Id
+    @Column(name = "address", columnDefinition = "uuid", updatable = false, nullable = false)
+    val address: UUID = UUID.randomUUID(),
```
abdallahFc: A wallet address is not a uuid. it’s a hash value generated from a private key make it as a string

wallet/src/main/kotlin/wallet/entity/Transaction.kt
```diff
@@ -0,0 +1,33 @@
...
+
+@Entity
+@Table(name = "transactions")
+data class Transaction(
+    @Id
+    @Column(columnDefinition = "uuid", updatable = false, nullable = false)
+    val id: UUID = UUID.randomUUID(),
+
+    @Column(nullable = false, updatable = false)
+    val timestamp: LocalDateTime = LocalDateTime.now(),
```
abdallahFc: rename to CreatedAt

wallet/src/main/kotlin/wallet/entity/Transaction.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    @Column(name = "sender_wallet_address", columnDefinition = "uuid", nullable = false, updatable = false)
+    val senderWalletAddress: UUID,
+
+    @Column(name = "receiver_wallet_address", columnDefinition = "uuid", nullable = false, updatable = false)
+    val receiverWalletAddress: UUID,
```
abdallahFc: senderAddress and receiverAddress are enough. there’s no need to include the word ‘wallet’ since the context is already clear also make it a string not uuid

abdallahFc (changes requested):

abdallahFc (approved):

wallet/src/main/kotlin/net/thechance/wallet/api/controller/WalletController.kt
```diff
@@ -1,12 +1,12 @@
...
+package net.thechance.wallet.api.controller
 
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.RestController
 
 @RestController
 class WalletController {
 
-    @GetMapping("/wallet/hello")
+    @GetMapping("/net/thechance/wallet/hello")
```
Karrar-Mohammed: why are we putting the path of this end point like this ?

Malak187: I didn't change it. Maybe when I created the two new packages (net/the chance), it automatically changed here for some reason, but I'll fix it.

Karrar-Mohammed (approved):
good job 👏

PR #8, opened 2025-09-15
[CHT-14]-Add basic structure  for chat feature 

raouf-kamel:
add basic structure
add Entity Contact


Joseph-Sameh-0 (dismissed):

chat/src/main/kotlin/chat/dto/.gitkeep
devfalah, line 1: This file should be placed in `chat/src/main/kotlin/net/thechance/chat/dto/.gitkeep`

devfalah: same for others

chat/src/main/kotlin/chat/entity/Contact.kt
```diff
@@ -0,0 +1,21 @@
+package net.thechance.chat.entity
+
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.util.UUID
+
+@Entity
+@Table(name = "contacts")
```
mosayed01: could we please specify the schema here

omer1998 (approved):

omer1998 (commented):
please raouf, update the folder structure to be aligned to what has been proposed by karrar

Joseph-Sameh-0:
add empty service and repository as a part of the structure

chat/src/main/kotlin/chat/mapper/.gitkeep
Joseph-Sameh-0, line 1: we don't need this folder

chat/src/main/kotlin/chat/entity/Contact.kt
```diff
@@ -0,0 +1,21 @@
...
+import java.util.UUID
+
+@Entity
+@Table(name = "contacts", schema = "chat")
+data class Contact(
+    @Id @Column(
+        columnDefinition = "uuid",
+        updatable = false,
+        nullable = false
+    ) val id: UUID = UUID.randomUUID(),
```
Karrar-Mohammed: you can put all the annotation parameter in one line for the readability

chat/src/main/kotlin/chat/entity/Contact.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.util.UUID
+
+@Entity
+@Table(name = "contacts", schema = "chat")
+data class Contact(
```
Karrar-Mohammed: why we have contact table ? why do we need to save the contact in the database ?

Joseph-Sameh-0: yes the user sync and save his contacts as a part of the story

omer1998: contact table here meant to save the contacts of the mena user in the database in order to enable the user to chat with them later on

chat/src/main/kotlin/chat/entity/Contact.kt
```diff
@@ -0,0 +1,21 @@
...
+    @Id @Column(
+        columnDefinition = "uuid",
+        updatable = false,
+        nullable = false
+    ) val id: UUID = UUID.randomUUID(),
+    @Column(nullable = false) val name: String,
```
Joseph-Sameh-0: it should be first name and last name not only one name

Joseph-Sameh-0 (approved):

chat/src/main/kotlin/chat/api/controller/ContactController.kt
```diff
@@ -0,0 +1,9 @@
+package net.thechance.chat.api.controller
+
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/contacts")
```
mosayed01:
```suggestion
@RequestMapping("/chat/contacts")
```

mosayed01 (approved):

monaayman123 (approved):

chat/src/main/kotlin/chat/entity/Contact.kt
devfalah: @Joseph-Sameh-0 @omer1998 theres no need to save contacts on the server side, the sync endpoint should work like this:
- client sends a list of contacts to the server
- server responds with the same list, but with an extra flag indicate has a Mena account or not
- based on this response, the client updates its local database

devfalah: imagine if we have more than 10K users and each user has 1K contacts saving them all would result +10M contacts in our database without any benefit

Joseph-Sameh-0: @devfalah 
so why is there a button in the app to resync the contacts?
this is not the requirements
the requirements is to sync and store the contacts 

image

chat/src/main/kotlin/chat/api/controller/ContactController.kt
Joseph-Sameh-0: why ?
this chat/ path will be reserved for the chats not the contacts

mosayed01: since it's multi module and multi responsibility so we need to categories each some endpoints into base path
@Karrar-Mohammed what do u think?

Karrar-Mohammed: @mosayed01 yes that's correct

Karrar-Mohammed (approved):
good job, just do one more thing make the packages of the module like this

net/thechance/chat/{the other packages}

you can check falah message on slack for more details

PR #9, opened 2025-09-15
[Feature/DKN-15] - Create dukan tables & repositories

devfalah:


dukan/src/main/kotlin/net/thechance/dukan/entity/Category.kt
```diff
@@ -0,0 +1,23 @@
...
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.util.UUID
+
+@Table(name = "categories")
+@Entity
+data class Category(
```
TarekIdrees: It would be more consistent if we change this to DukanCategory to follow the same convenient of the rest tables.

dukan/src/main/kotlin/net/thechance/dukan/entity/Category.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.util.UUID
+
+@Table(name = "categories")
```
TarekIdrees: Same as well here dukan_categories

TarekIdrees (changes requested):

dukan/src/main/kotlin/net/thechance/dukan/entity/Category.kt
devfalah: done, thank you

dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
```diff
@@ -0,0 +1,52 @@
...
+    val ownerId: String,
+    @Column(name = "name", nullable = false, unique = true)
+    val name: String,
+    @ManyToMany
+    @JoinTable(
+        name = "dukan_categories",
+        joinColumns = [JoinColumn(name = "dukan_id")],
+        inverseJoinColumns = [JoinColumn(name = "category_id")]
+    )
+    val categories: List<Category>,
```
haideraljawasim: I suggest we use here a Set here  instead of list since the categories should be unique

devfalah: Nice catch, fixed

haideraljawasim (approved):

ElsayedMagdy122 (approved):

TarekIdrees (approved):

dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
```diff
@@ -0,0 +1,52 @@
+package net.thechance.dukan.entity
+
+import jakarta.persistence.*
+import java.util.*
+
+@Table(name = "dukans")
```
Karrar-Mohammed: you need to specify the schema

dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
```diff
@@ -0,0 +1,52 @@
...
+@Entity
+data class Dukan(
+    @Id
+    @GeneratedValue(strategy = GenerationType.UUID)
```
Karrar-Mohammed, line 10: is this working for you ? and generating uuid as expected ?

dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
```diff
@@ -0,0 +1,52 @@
...
+    val id: UUID = UUID.randomUUID(),
+    @Column(name = "owner_id", nullable = false, unique = true)
+    val ownerId: String,
```
Karrar-Mohammed: i think this will need to be a foreign key to reference a the owner in the owner table, and it's type should be the same type of the owner primary key, so if the owner will be the user from the user table then it's type should be UUID

dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
```diff
@@ -0,0 +1,52 @@
...
+    @Column(name = "name", nullable = false, unique = true)
+    val name: String,
+    @ManyToMany
+    @JoinTable(
+        name = "dukan_categories_junction",
```
Karrar-Mohammed, line 18: do we need the junction at the end of the table name ? do you think just "dukan_categories" will be enough ?

dukan/src/main/kotlin/net/thechance/dukan/entity/DukanCategory.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import jakarta.persistence.Column
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.util.UUID
+
+@Table(name = "dukan_categories")
```
Karrar-Mohammed: same here i think you need to add the schema

dukan/src/main/kotlin/net/thechance/dukan/repository/CategoryRepository.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.dukan.repository
+
+import net.thechance.dukan.entity.DukanCategory
+import org.springframework.data.jpa.repository.JpaRepository
+import java.util.UUID
+
+interface CategoryRepository : JpaRepository<DukanCategory, UUID>
```
Karrar-Mohammed: do you think if we name it DukanCategoryRepository will be more clear ?

dukan/src/main/kotlin/net/thechance/dukan/repository/ColorRepository.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.dukan.repository
+
+import net.thechance.dukan.entity.DukanColor
+import org.springframework.data.jpa.repository.JpaRepository
+import java.util.UUID
+
+interface ColorRepository: JpaRepository<DukanColor, UUID>
```
Karrar-Mohammed: same here i suggest to name it DukanColorRespository so it can be easier to understand it's responsibility

dukan/src/main/kotlin/net/thechance/dukan/entity/DukanColor.kt
```diff
@@ -0,0 +1,19 @@
...
+import jakarta.persistence.Id
+import jakarta.persistence.Table
+import java.util.UUID
+
+@Table(name = "dukan_colors")
+@Entity
+data class DukanColor(
+    @Id
+    @GeneratedValue(strategy = GenerationType.UUID)
+    val id: UUID = UUID.randomUUID(),
```
Karrar-Mohammed, line 16: the colors are simple thing, i suggest to make the id just a long, what do you think ?

dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
devfalah: done

dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
devfalah: Yes it works for me

dukan/src/main/kotlin/net/thechance/dukan/entity/DukanCategory.kt
devfalah: done

dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
devfalah: `dukan_categories` table name is already used by the DukanCategory entity

dukan/src/main/kotlin/net/thechance/dukan/repository/CategoryRepository.kt
devfalah: you are right, I have renamed to DukanCategoryRepository

dukan/src/main/kotlin/net/thechance/dukan/repository/ColorRepository.kt
devfalah: done

dukan/src/main/kotlin/net/thechance/dukan/entity/DukanColor.kt
devfalah: I prefer to keep the IDs type consistent across entities, and since the number of colors is small I don’t think using UUIDs will be an issue

Karrar-Mohammed (approved):

PR #12, opened 2025-09-16
#DKN-17 Get available styles

TarekIdrees:
- This PR mainly focuses on getting all styles of the Dukan entity 
- Endpoint: `/dukan/styles`
- Expected returned body : 
```
 "styles": [
        "WIDE_IMAGE",
        "SMALL_IMAGE",
        "NO_IMAGE"
    ]
```


dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
```diff
@@ -1,13 +1,16 @@
...
 import org.springframework.web.bind.annotation.RestController
 
 @RestController
-class DukanController {
+@RequestMapping("/dukan")
+class DukanController(
+    private val dukanService: DukanService
+) {
+    @GetMapping("/styles")
+    fun getAllStyles () = dukanService.getAllStyles()
```
devfalah: I think it would be better to return `ResponseEntity<DukanStyleResponse>` here

TarekIdrees: Nicely done.
Please mark this as resolved once you are convinced of the new commit.

mostafa-n3ma (approved):
I don't really have any comments here every thing look nice , thank you

haideraljawasim (approved):

ElsayedMagdy122 (approved):

ElsayedMagdy122:
Great Work Tarek 👏

devfalah (approved):

dukan/src/main/kotlin/net/thechance/dukan/mapper/DukanCategroyMapper.kt
```diff
@@ -0,0 +1,8 @@
+package net.thechance.dukan.mapper
+
+import net.thechance.dukan.api.dto.DukanStyleResponse
+import net.thechance.dukan.entity.Dukan
+import kotlin.enums.EnumEntries
+
+fun EnumEntries<Dukan.Style>.toDukanStyleResponse(): DukanStyleResponse =
+    DukanStyleResponse(styles = this.map { it.name })
```
Karrar-Mohammed, line 8: i think you can make the styles list of type the enum class like

List<Dukan.Style>

then the json parser will automatically map it to json, so we don't need to write a mapper

this is just a suggetion and i'm also ok with keeping this mapper

app/src/main/resources/schema.sql
Karrar-Mohammed, line 1: i don't think we need that, because we should use liquibase instead

Karrar-Mohammed:
i don't have any issue with this PR only we need to remove the query for creating the schema

dukan/src/main/kotlin/net/thechance/dukan/mapper/DukanCategroyMapper.kt
TarekIdrees: To make it consistent with the rest of DukanStyleResponse, DukanCategroyResponse, etc

app/src/main/resources/schema.sql
TarekIdrees: Nicely done 
I have added the file to enable running the configuration locally.

devfalah (approved):

Karrar-Mohammed (approved):

PR #13, opened 2025-09-16
#DKN-20 Get available categories

TarekIdrees:
- This PR mainly focuses on getting all categories of the Dukan entity
- Endpoint: /dukan/categories
- Expected returned body :

```
[
    "categories" : {
        "id": "cd89d74c-de29-477b-87bd-6d1bb8808127",
        "icon": "icon_url 1",
        "title": "arabic_title 1",
    },
    {
        "id": "840706c7-5d80-4464-9a17-780a1dd1751b",
        "icon": "icon_url 2",
        "title": "arabic_title 2"
    }
]
```


dukan/src/main/kotlin/net/thechance/dukan/api/dto/CategoryDto.kt
```diff
@@ -0,0 +1,8 @@
+package net.thechance.dukan.api.dto
+
+data class CategoryDto(
+    val id: String,
+    val icon: String,
+    val arabicTitle: String,
+    val englishTitle: String,
```
devfalah: We can return title to the client and based on language header we return either english or arabic what do you think?

```suggestion
    val title: String,
```

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
```diff
@@ -1,13 +1,19 @@
...
-class DukanController {
+@RequestMapping("/dukan")
+class DukanController(
+    private val dukanService: DukanService
+) {
+    @GetMapping("/styles")
+    fun getAllStyles () = dukanService.getAllStyles()
+
+    @GetMapping("/categories")
+    fun getAllCategories () = dukanService.getAllCategories()
```
devfalah: It would be better to return `ResponseEntity<DukanCategoriesResponse>`

`data class DukanCategoriesResponse(val categories: List<CategoryDto>)`

dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
```diff
@@ -0,0 +1,24 @@
...
+    private val dukanRepository: DukanRepository,
+    private val dukanColorRepository: DukanColorRepository,
+    private val dukanCategoryRepository: DukanCategoryRepository
+) {
+    fun getAllStyles() = Dukan.Style.entries
+
+    fun getAllCategories(): List<CategoryDto> {
+        return  dukanCategoryRepository.findAll().let { categories ->
+            categories.map { category -> category.toDto() }
+        }
```
devfalah: In my opinion DTOs related to the `api layer (controller)` I think our service should deal with entities and the mapping should happen in the controller

dukan/src/main/kotlin/net/thechance/dukan/api/dto/CategoryDto.kt
TarekIdrees: make sense to me

dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
TarekIdrees: Nicely done

I assumed most of the mapping would be handled within the service so the controller could remain isolated and focus solely on processing the HTTP request.

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
TarekIdrees: Nicely done

dukan/src/main/kotlin/net/thechance/dukan/api/dto/CategoryDto.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.dukan.api.dto
+
+data class CategoryDto(
```
haideraljawasim: doesn't this represent the `DukanCategory` class? I think we should keep the same naming convention and name it `DukanCategoryDto`

dukan/src/main/kotlin/net/thechance/dukan/api/dto/CategoryResponse.kt
```diff
@@ -0,0 +1,5 @@
+package net.thechance.dukan.api.dto
+
+data class CategoryResponse(
```
haideraljawasim: same here i suggest `DukanCategoryResponse`

haideraljawasim (approved):

ElsayedMagdy122 (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/dto/CategoryDto.kt
TarekIdrees: Agree

dukan/src/main/kotlin/net/thechance/dukan/api/dto/CategoryResponse.kt
TarekIdrees: Agree

devfalah (approved):

app/src/main/resources/schema.sql
Karrar-Mohammed, line 1: i don't think this is how we should handle the database migration, we should wait for the liquibase

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
```diff
@@ -1,13 +1,28 @@
...
 import org.springframework.web.bind.annotation.RestController
 
 @RestController
-class DukanController {
+@RequestMapping("/dukan")
+class DukanController(
+    private val dukanService: DukanService
+) {
+    @GetMapping("/styles")
+    fun getAllStyles() = dukanService.getAllStyles()
```
Karrar-Mohammed, line 24: what is this styles ? what it's return type and why it's not using the ResponseEntity ?

Karrar-Mohammed (changes requested):
good job, just we need to fix the styles end point return type, and remove writing query directly

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
TarekIdrees: These are entries retrieved from an enum class where there is no need to execute a query in order to get the values

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
```diff
@@ -1,13 +1,28 @@
...
-        return "Hello Dukan World!"
+    @GetMapping("/categories")
+    fun getAllCategories(): ResponseEntity<DukanCategoryResponse> {
+        val categories = dukanService.getAllCategories().let { categories ->
+            // Todo replace default Arabic with the extracted language from the header
+            categories.map { category -> category.toDto(DukanLanguage.ARABIC) }
+        }
+        return ResponseEntity.ok(DukanCategoryResponse(categories))
     }
 }
```
TarekIdrees: @Karrar-Mohammed 
here where the main logic happens

app/src/main/resources/schema.sql
TarekIdrees: I know where I have to add this in order to be able to run the module locally because the main entities marked with a dukan schema.

Karrar-Mohammed: ok if you need it for you local database then execute it directly on your local database, do not push it or commit because it may cause some issues on the production

Karrar-Mohammed (approved):

PR #14, opened 2025-09-16
[Feature/DKN-15] Add Dukan name checking API

devfalah:
This PR implements the Dukan name checking API endpoint as part of the Dukan creation flow.
Endpoint:
```
GET /dukan/available?name=MyDukan
``` 

```
{
  "available": true //indicates that the name is available for use
}
```

### DIFF
--- a/dukan/src/main/kotlin/net/thechance/dukan/repository/DukanRepository.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/repository/DukanRepository.kt
@@ -2,6 +2,10 @@ package net.thechance.dukan.repository
 
 import net.thechance.dukan.entity.Dukan
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
 import java.util.UUID
 
-interface DukanRepository: JpaRepository<Dukan, UUID>
+@Repository
+interface DukanRepository : JpaRepository<Dukan, UUID> {
+    fun existsByName(name: String): Boolean
+}
--- a/dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
@@ -7,7 +7,5 @@ import org.springframework.stereotype.Service
 class DukanService(
     private val dukanRepository: DukanRepository,
 ) {
-    fun doesNameExist(name: String): Boolean {
-        return dukanRepository.existsByName(name)
-    }
+    fun isNameExist(name: String): Boolean = dukanRepository.existsByName(name)
 }
--- a/dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
@@ -1,0 +1,13 @@
+package net.thechance.dukan.service
+
+import net.thechance.dukan.repository.DukanRepository
+import org.springframework.stereotype.Service
+
+@Service
+class DukanService(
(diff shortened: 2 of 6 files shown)


dukan/src/main/kotlin/net/thechance/dukan/repository/DukanRepository.kt
```diff
@@ -2,6 +2,10 @@ package net.thechance.dukan.repository
...
 import net.thechance.dukan.entity.Dukan
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
 import java.util.UUID
 
-interface DukanRepository: JpaRepository<Dukan, UUID>
\ No newline at end of file
+@Repository
+interface DukanRepository : JpaRepository<Dukan, UUID> {
+    fun existsByName(name: String): Boolean
```
ElsayedMagdy122, line 10: I suggest renaming it to checkNameIsExists or DukanName to make the name more clear and indicative of the return type.

ElsayedMagdy122 (approved):

ElsayedMagdy122:
Good Work Falah 👏

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanRepository.kt
devfalah: Thanks for your suggestion, for repository methods we should follow JPA query method naming so that JPA can understand them and convert them into queries

https://docs.spring.io/spring-data/jpa/reference/jpa/query-methods.html

ElsayedMagdy122: Thanks for the clarification, I wouldn't have known that.🤎

dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
```diff
@@ -0,0 +1,11 @@
+package net.thechance.dukan.service
+
+import net.thechance.dukan.repository.DukanRepository
+import org.springframework.stereotype.Service
+
+@Service
+class DukanService(
+    private val dukanRepository: DukanRepository,
+) {
+    fun isNameExist(name: String): Boolean = dukanRepository.existsByName(name)
```
mostafa-n3ma: I think in the service it might be clearer to reverse the logic and rename this method to isDukanNameAvailable().
	•	If the name already exists, we should return false, showing that it’s taken and not available.
	•	If it doesn’t exist, we return true (200 OK), meaning the name is available for use.

This way the service method directly reflects the business intent: preventing users from creating duplicated dukan name. What do you think?

TarekIdrees: I would say isDukanExist will be more clear? Same as will for the parameter dukanName: String

TarekIdrees (changes requested):

dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
devfalah: Good point, thanks for pointing it out, done

devfalah: I renamed it to `isDukanNameAvailable` but for the parameter name I think name is enough since dukan is already mentioned in the method name

mostafa-n3ma (approved):

haideraljawasim (approved):

Karrar-Mohammed (approved):

TarekIdrees (approved):

PR #16, opened 2025-09-17
Feature/FTH-13 Save aya to bookmark

FaresM0hamed:
## What’s included in this PR
- Implemented `saveBookmark` endpoint  
- Added validation for the request body  
- Created unit tests for the business logic  

---

### Endpoint
```
POST http://localhost:8080/faith/ayah/bookmark
```
---

### Request Body
```json
{
  "surahId": 1,
  "ayahNumber": 3
}

```
### Response
```json
{
    "id": "10",
    "surahId": 1,
    "ayahNumber": 3,
    "createdAt": "2025-09-17T00:38:21.583169Z"
}

### CODE REVIEW

**Review by marwanqashwa — APPROVED** [review]

File: faith/src/test/kotlin/net/thechance/faith/service/BookmarkServiceTest.kt [hunk 1]
```diff
@@ -0,0 +1,35 @@
...
+    private val bookmarkService = BookmarkService(bookmarkRepository)
+
+    @Test
+    fun `saveBookmark should save bookmark and return saved entity when it called`() {
+        every { bookmarkRepository.save(bookmark) } returns bookmark
+
+        val result = bookmarkService.saveBookmark(bookmark)
+
+        assertThat(result).isEqualTo(bookmark)
+    }
```
yusufnasserdev: Can we add given/when/then comments here as a standard test writing practice?

faith/src/main/kotlin/net/thechance/faith/api/dto/bookmark/BookmarkRequest.kt
```diff
@@ -0,0 +1,12 @@
+package net.thechance.faith.api.dto.bookmark
+
+import jakarta.validation.constraints.Min
+import org.hibernate.validator.constraints.Range
+
+data class BookmarkRequest(
+    @field:Min(value = 1, message = "Surah ID must be at least 1")
```
yusufnasserdev: Why not set the max to `114` as it would never be more?

faith/src/main/kotlin/net/thechance/faith/api/dto/bookmark/BookmarkRequest.kt
```diff
@@ -0,0 +1,12 @@
...
+    val surahId: Int,
+
+    @field:Range(min = 1, max = 286, message = "Ayah number must be between 1 and 286")
```
yusufnasserdev: Doesn't the max vary from one Surah to Another?, should it be set here?

faith/src/main/kotlin/net/thechance/faith/api/controller/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,23 @@
+package net.thechance.faith.api.controller.exception
+
+import org.springframework.http.HttpStatus
+import org.springframework.web.bind.MethodArgumentNotValidException
+import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.bind.annotation.ResponseStatus
+import org.springframework.web.bind.annotation.RestControllerAdvice
+
+@RestControllerAdvice
+class GlobalExceptionHandler {
```
yusufnasserdev: Why is it called `Global`?

faith/src/test/kotlin/net/thechance/faith/service/BookmarkServiceTest.kt
FaresM0hamed: We could separate this sections by empty line , It more clean and make the test fun small

faith/src/main/kotlin/net/thechance/faith/api/dto/bookmark/BookmarkRequest.kt
FaresM0hamed: An ID is a variable thing  we don’t know how it will look; for example, it could be something like `27523`. If it were a `surahNumber`, then I would agree with you

faith/src/main/kotlin/net/thechance/faith/api/dto/bookmark/BookmarkRequest.kt
FaresM0hamed: Yes, but that’s not the main point here. The maximum number of ayat in the Qur’an is `286`` (in Al-Baqarah)`, so it doesn’t make sense to receive something like `ayahNumber = 300` from the user. At the same time, I’m not adding validation for each surah and its ayat ,Because that’s not my responsibility in Backend .

faith/src/main/kotlin/net/thechance/faith/api/controller/exception/GlobalExceptionHandler.kt
FaresM0hamed: In Spring Boot, under the hood, it throws a general exceptions rather than a custom one that I define, which is why it’s conventionally called global. Also, `@RestControllerAdvice `can catch exceptions thrown from any layer.

marwanqashwa: what about making it specific ?
 @RestControllerAdvice(basePackages  =["net.thechance.faith.api.controller"])

FaresM0hamed: It's Already Global for only our faith module , Not all project , so I think no need for this line 
`(basePackages =["net.thechance.faith.api.controller"])`

marwanqashwa: if you make sure that it will not catch anything from outside then it's ok, but i am not sure

faith/src/main/kotlin/net/thechance/faith/api/dto/bookmark/BookmarkRequest.kt
yusufnasserdev: Okay, but the message refers to it as `Surah ID`.

faith/src/main/kotlin/net/thechance/faith/api/dto/bookmark/BookmarkRequest.kt
yusufnasserdev: What's the point of checking at all, if we're not gonna depend on it for all Surahs?, Why not let them all be handled the same way?

faith/src/test/kotlin/net/thechance/faith/service/BookmarkServiceTest.kt
yusufnasserdev: Yeah, but when we're gonna have a test case not this small, we gonna want something to separate the sections, like these comments, it's fine now, but won't be fine then.

faith/build.gradle.kts
```diff
@@ -3,9 +3,18 @@ plugins {
...
+    kotlin("plugin.jpa")
 }
 
 dependencies {
-    testImplementation(kotlin("test"))
+    implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+    implementation("org.springframework.boot:spring-boot-starter-security")
+    implementation("org.springframework.boot:spring-boot-starter-validation")
     implementation("org.springframework.boot:spring-boot-starter-web")
+    runtimeOnly("org.postgresql:postgresql")
```
Karrar-Mohammed: you don't need this in the faith module only the jpa dependency is enough

faith/src/main/kotlin/net/thechance/faith/api/controller/BookmarkController.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.web.bind.annotation.RestController
+import java.util.*
+
+@RestController
+@RequestMapping("faith/bookmark")
+class BookmarkController(
+    private val bookmarkService: BookmarkService
+) {
+    @PostMapping("/save")
+    fun saveBookmark(
```
Karrar-Mohammed: i suggest to make the name of the function like bookmarkAya or saveAya

faith/src/main/kotlin/net/thechance/faith/api/controller/BookmarkController.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.util.*
+
+@RestController
+@RequestMapping("faith/bookmark")
+class BookmarkController(
+    private val bookmarkService: BookmarkService
+) {
+    @PostMapping("/save")
```
Karrar-Mohammed: also the path of the end point make like /aya/bookmark

faith/src/main/kotlin/net/thechance/faith/api/controller/BookmarkController.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.security.core.annotation.AuthenticationPrincipal
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.util.*
+
+@RestController
+@RequestMapping("faith/bookmark")
+class BookmarkController(
```
Karrar-Mohammed: why we call this controller bookmark ? will it be responsible for all the book marks in the module ? i think we only have to book mark aya so i changing the controller to be responsible of all the aya responsibility it's better and one of the functionality is the bookmark

faith/src/main/kotlin/net/thechance/faith/api/controller/exception/GlobalExceptionHandler.kt
Karrar-Mohammed: actually in this way it will catch from all the project not only the faith module

faith/src/main/kotlin/net/thechance/faith/api/controller/exception/GlobalExceptionHandler.kt
Karrar-Mohammed, line 1: i suggest to remove this error handler because other team already handle it for the all the project

faith/src/main/kotlin/net/thechance/faith/api/controller/exception/BaseErrorResponse.kt
Karrar-Mohammed, line 1: i suggest to remove it other team already did it

faith/src/main/kotlin/net/thechance/faith/entity/Bookmark.kt
```diff
@@ -0,0 +1,28 @@
...
+    @Column(nullable = false)
+    val surahId: Int,
+
+    @Column(nullable = false)
+    val ayahNumber: Int,
+
+    @CreationTimestamp
+    @Column(nullable = false, updatable = false)
+    val createdAt: Instant = Instant.now(),
+)
```
Karrar-Mohammed: the entity name is bookmark, but you explicitly have mentioned the surahId and ayah number, so it's not a general book mark it's ayah bookmark, so i suggest to rename it to AyahBookmark

faith/src/test/kotlin/net/thechance/faith/service/BookmarkServiceTest.kt
Karrar-Mohammed, line 1: it's good to write unit test, but we decided to write unit test for only the logic or if we have calculation, so for this case i don't think it's worth to spend time writing unit test for this simple functionality,

faith/src/main/kotlin/net/thechance/faith/service/BookmarkService.kt
```diff
@@ -0,0 +1,14 @@
...
+import net.thechance.faith.repository.BookmarkRepository
+import org.springframework.stereotype.Service
+
+@Service
+class BookmarkService(
+    private val bookmarkRepository: BookmarkRepository
+) {
+    fun saveBookmark(bookmark: Bookmark): Bookmark {
+        return bookmarkRepository.save(bookmark)
+    }
```
Karrar-Mohammed: if we will not have any logic in this service in the future, then i suggest to remove it and use the repository directly in the controller

FaresM0hamed: We still have other crud operations like delete and get , So in delete as example we still make some logic to check if this id exist or this userId exist then delete , so what about make all our crud in service and depend on it on controller in this case ?

FaresM0hamed:
All your comments has been resolved , So Please check it again 
@Karrar-Mohammed

yusufnasserdev (approved):
Great work!

faith/src/main/kotlin/net/thechance/faith/api/controller/exception/GlobalExceptionHandler.kt
marwanqashwa: So we can add it to our module with specific path to handle only our exceptions
@Karrar-Mohammed

marwanqashwa (approved):

faith/src/main/kotlin/net/thechance/faith/service/BookmarkService.kt
Karrar-Mohammed: if we will have logic in the future then we can keep the service

faith/build.gradle.kts
```diff
@@ -3,9 +3,17 @@ plugins {
...
 dependencies {
-    testImplementation(kotlin("test"))
+    implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+    implementation("org.springframework.boot:spring-boot-starter-security")
+    implementation("org.springframework.boot:spring-boot-starter-validation")
     implementation("org.springframework.boot:spring-boot-starter-web")
+
+    testImplementation(kotlin("test"))
+    testImplementation("io.mockk:mockk:1.13.12")
+    testImplementation("com.google.truth:truth:1.4.4")
```
Karrar-Mohammed, line 17: if you decided to remove the test then i think you can also remove the dependency

Karrar-Mohammed (approved):
great job 👏

PR #17, opened 2025-09-17
Feature/CHT-8-create-endpoint-for-sync-contacts

monaayman123:
This PR introduces a new feature for syncing user contacts in the chat module.
Users can send their contact list to the backend, which will update, insert, or remove contacts 
- Implemented ContactService to handle syncing user contacts
- Added ContactController with /chat/contacts/sync endpoint
- Created ContactRepository with custom queries for syncing logic
- Introduced ContactRequest DTO and mapper to Contact entity


Joseph-Sameh-0 (approved):

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
```diff
@@ -1,7 +1,35 @@
...
+
+        val contactsToSave = uniqueContactRequests.map { request ->
+            existingContactsMap[request.phoneNumber]?.copy(
+                firstName = request.firstName,
+                lastName = request.lastName
+            ) ?: request.toContact(userId)
+        }
+
+        contactRepository.deleteAllByUserIdAndPhoneNumberNotIn(userId, requestedPhoneNumbers)
+        contactRepository.saveAll(contactsToSave)
```
ZainabAlnajjar: I don't get why you are mapping the `uniqueContactRequests` to `existingContactsMap` and not just map it directly with `request.toContact(userId)` then save it, since you are deleting the old contacts before saving, could you clarify?

Joseph-Sameh-0: this mapping updates the contacts that already exist and add that don't exist with new Id
if we just do `request.toContact(userId)` the existing contacts will have new id and added again

chat/src/main/kotlin/net/thechance/chat/entity/Contact.kt
```diff
@@ -4,15 +4,22 @@ import jakarta.persistence.Column
...
+import jakarta.persistence.Index
+import jakarta.persistence.UniqueConstraint
 import java.util.UUID
 
 @Entity
-@Table(name = "contacts", schema = "chat")
+@Table(
+    name = "contacts",
+    schema = "chat",
+    indexes = [Index(name = "idx_contacts_user_id", columnList = "userId")],
```
raouf-kamel: I don't think we currently need to use indexes.

raouf-kamel (approved):

chat/src/main/kotlin/net/thechance/chat/entity/Contact.kt
Joseph-Sameh-0: the index will increase the performance of get by user Id

chat/src/main/kotlin/net/thechance/chat/api/controller/ContactController.kt
```diff
@@ -1,9 +1,28 @@
...
+    private val contactService: ContactService
+) {
+
+    @PostMapping("/sync")
+    fun syncContacts(
+        @RequestBody @Valid contacts: List<ContactRequest>,
+        @AuthenticationPrincipal userId: UUID
+    ): ResponseEntity<String> {
+        contactService.syncContacts(userId, contacts)
+        return ResponseEntity.ok("Contacts synced successfully")
```
Karrar-Mohammed: if that the case and we don't need to return anything, then just return 201 no content, without any body

chat/src/main/kotlin/net/thechance/chat/entity/Contact.kt
Karrar-Mohammed: remove the indexes for now, when we will face performance issue then we will add it, avoid early optimization

chat/src/main/kotlin/net/thechance/chat/entity/Contact.kt
```diff
@@ -4,15 +4,21 @@ import jakarta.persistence.Column
...
+    uniqueConstraints = [UniqueConstraint(columnNames = ["userId", "phoneNumber"])]
+)
 data class Contact(
     @Id @Column(columnDefinition = "uuid", updatable = false, nullable = false)
     val id: UUID = UUID.randomUUID(),
     @Column(nullable = false) val firstName: String,
     @Column(nullable = false) val lastName: String,
     @Column(nullable = false) val phoneNumber: String,
-    @Column(nullable = false) val userId: UUID,
+    @Column(nullable = false) val userId: UUID
```
Karrar-Mohammed: i think this is the user id of the contact owner and not the user id of the phoneNumber owner, so in that case change the name of this Parameter to contactOwnerId so it can be clear and easier to understand

chat/src/main/kotlin/net/thechance/chat/repository/ContactRepository.kt
```diff
@@ -4,4 +4,7 @@ import net.thechance.chat.entity.Contact
 import org.springframework.data.jpa.repository.JpaRepository
 import java.util.UUID
 
-interface ContactRepository: JpaRepository<Contact, UUID>
\ No newline at end of file
+interface ContactRepository : JpaRepository<Contact, UUID> {
+    fun findAllByUserIdAndPhoneNumberIn(userId: UUID, phoneNumbers: List<String>): List<Contact>
```
Karrar-Mohammed: we should not return a full list, this should be paginated

chat/src/main/kotlin/net/thechance/chat/repository/ContactRepository.kt
```diff
@@ -4,4 +4,7 @@ import net.thechance.chat.entity.Contact
...
+    fun deleteAllByUserIdAndPhoneNumberNotIn(userId: UUID, requestedPhoneNumbers: List<String>)
```
Karrar-Mohammed: why do we need this function ? and why it's using the NotIn ?

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
```diff
@@ -1,7 +1,35 @@
...
+) {
+    fun syncContacts(userId: UUID, contactRequests: List<ContactRequest>) {
+
+        val uniqueContactRequests = contactRequests
+            .groupBy { it.phoneNumber }
+            .map { it.value.last() }
+        val requestedPhoneNumbers = uniqueContactRequests.map { it.phoneNumber }
+
+        val existingContactsMap = contactRepository
+            .findAllByUserIdAndPhoneNumberIn(userId, requestedPhoneNumbers)
```
Karrar-Mohammed: this is not a good approach because if the list is too big it might fill all the ram of the server

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
```diff
@@ -1,7 +1,35 @@
...
+import net.thechance.chat.api.dto.toContact
 import org.springframework.stereotype.Service
+import java.util.UUID
 
 @Service
-class ContactService {
+class ContactService(
+    private val contactRepository: ContactRepository,
+) {
+    fun syncContacts(userId: UUID, contactRequests: List<ContactRequest>) {
```
Karrar-Mohammed: it's better to not pass the request data class to the service

chat/src/main/kotlin/net/thechance/chat/repository/ContactRepository.kt
Joseph-Sameh-0: when the user sync the contacts we should delete the contacts that not in his contacts 
and update the existing and add the new 
so I should do findAllByUserIdAndPhoneNumberIn to update
and do deleteAllByUserIdAndPhoneNumberNotIn to delete the contact that he don't have

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
Joseph-Sameh-0: The user send all his contacts so what do you recommend?

chat/src/main/kotlin/net/thechance/chat/repository/ContactRepository.kt
Joseph-Sameh-0: we don't return this to the user 
we use it in the service to complete syncing user contacts

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
Joseph-Sameh-0: The other approach is to delete all and save all again
it's worst than my approach

chat/src/main/kotlin/net/thechance/chat/repository/ContactRepository.kt
Joseph-Sameh-0: do you have a better solution?

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
Joseph-Sameh-0: It's really heavy operation and the user should wait because it takes time

Karrar-Mohammed: what i mean is that we should not get all the contacts at once because it might be too big and crash the server, so it's better to be paginated

Joseph-Sameh-0: If I make it paginated how I will update all the contacts?
and know what is the deleted contacts?

local.properties
Karrar-Mohammed, line 1: why do we have local.properties here in the backend ?

Joseph-Sameh-0: It's uploaded by mistake

Karrar-Mohammed (approved):

PR #18, opened 2025-09-17
Feature/ TRN-10-add implementation of delete trend functionality

ahmeddsamii:
- Add trend `Reel` entity
- implement `ReelsController` delte function
- Add the logic inside the `ReelsService`
- Add `ReelsRepository`
- Add `BaseErrorResponse` class
- Add `GlobalExceptionHandler`


trends/src/main/kotlin/net/thechance/trends/entity/Reel.kt
```diff
@@ -0,0 +1,36 @@
...
+import java.util.*
+
+@Table(name = "reels", schema = "trends")
+@Entity
+data class Reel(
+    @Id
+    @GeneratedValue(strategy = GenerationType.UUID)
+    val id: UUID = UUID.randomUUID(),
+    @Column(name = "user_id", nullable = false)
+    val userId: UUID,
```
mohamedd-hassan: i would suggest naming this `publisherUserId` for clarity

ahmeddsamii: I think it's clear, there will not be any other user here except the publisher

ahmeddsamii: U are right, but i think ownerId will be better

mohamedd-hassan: yes even better

mohamedd-hassan (approved):

trends/src/main/kotlin/net/thechance/trends/entity/Reel.kt
```diff
@@ -0,0 +1,35 @@
...
+import jakarta.persistence.*
+import java.time.LocalDateTime
+import java.util.*
+
+@Table(name = "reels", schema = "trends")
+@Entity
+data class Reel(
+    @Id
+    @GeneratedValue(strategy = GenerationType.UUID)
+    val id: UUID = UUID.randomUUID(),
```
mohamedd-hassan, line 12: generated value should not be used for UUID right?
karrar defined it like this if i remember correctly
`@Column(columnDefinition = "uuid", updatable = false, nullable = false)`

ahmeddsamii: I don't know but when i tested it worked as expected

trends/src/main/kotlin/net/thechance/trends/api/controller/ReelsController.kt
```diff
@@ -0,0 +1,38 @@
...
+
+@RestController
+@RequestMapping("/reels")
+class ReelsController(
+    private val reelsService: ReelsService
+) {
+
+    @DeleteMapping("/{id}")
+    fun deleteReelById(@PathVariable id: UUID): ResponseEntity<Any> {
+        return runCatching {
```
iZyadNasser: can't you extract the error handling into a separate function for reusability?

trends/src/main/kotlin/net/thechance/trends/entity/Reel.kt
```diff
@@ -0,0 +1,35 @@
+package net.thechance.trends.entity
+
+import jakarta.persistence.*
+import java.time.LocalDateTime
+import java.util.*
+
+@Table(name = "reels", schema = "trends")
+@Entity
+data class Reel(
```
iZyadNasser, line 9: do we need all these properties in the current story? can't we remove some of them for simplicity and add them when we need them?

trends/src/main/kotlin/net/thechance/trends/service/ReelsService.kt
```diff
@@ -0,0 +1,24 @@
...
+import java.util.*
+
+@Service
+class ReelsService(
+    private val reelsRepository: ReelsRepository,
+) {
+
+    fun deleteReelById(id: UUID) {
+        val reel = reelsRepository.findById(id)
+            .orElseThrow { NoSuchElementException("Reel with id $id not found") }
```
iZyadNasser: same idea, try to extract error handling in another util function so we can all use it

ahmeddsamii: This is just throwing an exception and i think we will not throw the same exception always

trends/src/main/kotlin/net/thechance/trends/entity/Reel.kt
ahmeddsamii: This is the whole table i guess and it's okay to make it like this to not update it again later and it will not cause any issue and already some of them has default values

ahmeddsamii: And yes it's required to be on the screen in this sprint

iZyadNasser (approved):

trends/src/main/kotlin/net/thechance/trends/api/controller/ReelsController.kt
FaresM0hamed, line 1: I recommend using @RestControllerAdvice to handle exceptions automatically. With it, you won’t need to manage exceptions manually. It’s also a cleaner approach since you can throw exceptions from any layer, and the advisor will catch them and return an error response with your exception message

trends/src/main/kotlin/net/thechance/trends/entity/Reel.kt
FaresM0hamed, line 1: You can remove the name parameter from each column since PostgreSQL automatically converts your fields to columns with snake_case names, even if they’re written in camelCase. I tried this, and it worked for me.

trends/src/main/kotlin/net/thechance/trends/api/controller/ReelsController.kt
ahmeddsamii: done

trends/src/main/kotlin/net/thechance/trends/service/ReelsService.kt
ahmeddsamii: done

trends/src/main/kotlin/net/thechance/trends/entity/Reel.kt
```diff
@@ -0,0 +1,35 @@
...
+    @Column(name = "owner_id", nullable = false)
+    val ownerId: UUID,
+    @Column(name = "thumbnail_url", nullable = false)
+    val thumbnailUrl: String,
+    @Column(name = "video_url", nullable = false)
+    val videoUrl: String,
+    @Column(name = "description", nullable = false)
+    val description: String,
+    @Column(name = "likes_count", nullable = false)
+    val likesCount: Int = 0,
```
FaresM0hamed, line 22: "I was thinking about this column, and I believe it would be more flexible to model it as a separate table, for example, Likes, containing `userId`, `createdAt`, and `emojiType`.

The benefit is that you can easily track who liked a video and when. If you also want to calculate the total number of likes, you could handle that in the service layer using something like a flyweight calculation. This approach reduces unnecessary writes to the database, since with a counter column you would need to update it every time a user likes a video.

What do you think?"

trends/src/main/kotlin/net/thechance/trends/service/ReelsService.kt
```diff
@@ -0,0 +1,24 @@
...
+
+        val currentUserId = SecurityContextHolder.getContext().authentication.principal as UUID
+        if (reel.ownerId != currentUserId) {
+            throw IllegalAccessException("You can only delete your own reels")
+        }
```
FaresM0hamed: I recommend passing userId from the controller to the service using `@AuthenticationPrincipal userId: UUID`

Also, regarding the error message: currently it tells the user they’re trying to delete something that doesn’t belong to them. I think this could be considered a security concern. Instead, it would be better to return a more generic message, such as 'Reel not found', without revealing that the reel belongs to someone else.

FaresM0hamed (commented):
Great Work Ahmed 👏

trends/src/main/kotlin/net/thechance/trends/entity/Reel.kt
ahmeddsamii: Thank you for the advice but it's not the responsibility of this user story but i 'll keep it in my mind in the next user story

trends/src/main/kotlin/net/thechance/trends/api/controller/ReelsController.kt
ahmeddsamii: Already i did it and thank you

FaresM0hamed (approved):

trends/src/main/kotlin/net/thechance/trends/api/controller/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,34 @@
...
+
+import net.thechance.trends.api.controller.exception.ReelNotFoundException
+import net.thechance.trends.api.dto.BaseErrorResponse
+import org.springframework.http.HttpStatus
+import org.springframework.web.bind.annotation.ExceptionHandler
+import org.springframework.web.bind.annotation.ResponseStatus
+import org.springframework.web.bind.annotation.RestControllerAdvice
+
+@RestControllerAdvice
+class GlobalExceptionHandler {
```
Karrar-Mohammed: make this class to handle only the chat module exceptions and change it's name to ChatExceptionHandler

trends/src/main/kotlin/net/thechance/trends/api/dto/BaseErrorResponse.kt
```diff
@@ -0,0 +1,9 @@
+package net.thechance.trends.api.dto
+
+import com.fasterxml.jackson.annotation.JsonInclude
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+    data class BaseErrorResponse(
```
Karrar-Mohammed: i suggest to delete this class and wait when we will merge the global exception handler, so we can use the same error response

Karrar-Mohammed (approved):
good job, just consider my comments regarding the exception handler

trends/src/main/kotlin/net/thechance/trends/api/controller/GlobalExceptionHandler.kt
ahmeddsamii: You mean like this?
@RestControllerAdvice(basePackages = ["net.thechance.trends.api.controller"])

ahmeddsamii: done

PR #19, opened 2025-09-17
Feature/TRN-26 setup application properties

mohamedd-hassan:
created `application-dev.properties` and `application-prod.properties`
created a `.env` file and added it to gitignore to set environment variables there
changed `Dockerfile` entry point to make start profile `prod`
made the jwt expiration variable depending on if we are running in dev or prod to make development easier
added logging for dev profile and made it just warn for prod
made `server.port` able to be set through the `.env` file with a default of `8080` if it was not set


Karrar-Mohammed:
great job, and thank you for your effort, but we will keep this PR open and merge it later because it need to adjust the deployment process and we don't have time to do it now

norhanadel367 (approved):

ahmednasserzaza (commented):
Are these changes the responsibility of the Trends team? I think they are related to the Infrastructure team

PR #20, opened 2025-09-17
add global exception handler

Ahmed7osny1:
- Add ApiErrorResponse data class for unified error responses
- Add ErrorCodes object with business-specific error codes  
- Add GlobalExceptionHandler for centralized exception handling
- Add custom exceptions grouped by domain:
  * AuthenticationExceptions (credentials, tokens)
  * ResourceExceptions (CRUD operations) 
  * ValidationExceptions (input validation)
- Support both custom error codes and HTTP status codes
- Include proper logging for different exception types


app/src/main/kotlin/net/thechance/app/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,58 @@
...
+        return createErrorResponse(ex, HttpStatus.BAD_REQUEST.value(), "Missing Required Field")
+    }
+
+    @ExceptionHandler(Exception::class)
+    fun handleGeneral(): ResponseEntity<ApiErrorResponse> {
+        val apiError = ApiErrorResponse(
+            status = HttpStatus.INTERNAL_SERVER_ERROR.value(),
+            message = "An unexpected error occurred while processing your request"
+        )
+        return ResponseEntity(apiError, HttpStatus.INTERNAL_SERVER_ERROR)
```
Karrar-Mohammed: why don't you here also you the createErrorResponse function ?

app/src/main/kotlin/net/thechance/app/exception/CustomAuthenticationEntryPoint.kt
Karrar-Mohammed, line 1: i suggest to put this class in the identity module

identity/src/main/kotlin/identity/security/JwtFilter.kt
```diff
@@ -40,11 +45,55 @@ class JwtFilter(
...
+
+    private fun sendErrorResponse(
+        response: HttpServletResponse,
+        apiError: ApiErrorResponse
+    ) {
+        response.contentType = "application/json"
+        response.status = HttpServletResponse.SC_UNAUTHORIZED
+
+        response.writer.write(objectMapper.writeValueAsString(apiError))
+    }
```
Karrar-Mohammed: it's better to move these functions outside this class, i think you can put it in the entryPoint class

app/src/main/kotlin/net/thechance/app/exception/GlobalExceptionHandler.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    @ExceptionHandler(ResourceNotFoundException::class)
+    fun handleResourceNotFound(ex: ResourceNotFoundException): ResponseEntity<ApiErrorResponse> {
+        return createErrorResponse(ex, HttpStatus.NOT_FOUND.value(), "Resource Not Found")
+    }
+
+    @ExceptionHandler(ResourceConflictException::class)
+    fun handleResourceConflict(ex: ResourceConflictException): ResponseEntity<ApiErrorResponse> {
+        return createErrorResponse(ex, HttpStatus.CONFLICT.value(), "Resource Conflict")
+    }
```
Karrar-Mohammed: why do we need to handle these exceptions ?

identity/src/main/kotlin/identity/security/JwtFilter.kt
Ahmed7osny1: I searched at this point and found:
- CustomAuthenticationEntryPoint implements AuthenticationEntryPoint, which Spring Security only calls automatically when an unauthenticated request hits a protected endpoint.
- JwtFilter runs before that happens. If the token has expired, it's being handled manually inside the filter.
What's your opinion?

app/src/main/kotlin/net/thechance/app/exception/GlobalExceptionHandler.kt
Ahmed7osny1: handleResourceConflict: used for duplicate entries or business rule violations 
handleResourceNotFound: used when data is missing

your opinion, remove these, and each module handles these if needed?

identity/src/main/kotlin/identity/security/JwtFilter.kt
Karrar-Mohammed: at least create a class which contain these functions to handle the jwt errors

app/src/main/kotlin/net/thechance/app/exception/GlobalExceptionHandler.kt
Karrar-Mohammed: i don't think we will need it so it's better to remove them

Karrar-Mohammed (approved):

Malak187 (approved):

MuhammedEdrees (approved):

PR #21, opened 2025-09-17
#DKN-19 Get Dukan statues

TarekIdrees:
- This PR mainly focuses on getting Dukan statues
- Endpoint: `dukan/statues`

- Expected returned body :
```

{
    "name": "Adidas",
    "statues": "APPROVED"
}
```

### DIFF
--- a/dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
@@ -28,9 +28,9 @@ class DukanController(
 
     @GetMapping("/statues")
     fun getDukanStatues(
-        @AuthenticationPrincipal ownerId: UUID,
+        @AuthenticationPrincipal userId: UUID,
     ): ResponseEntity<DukanStatuesResponse> {
-        val dukan = dukanService.getDukanByOwnerId(ownerId)
+        val dukan = dukanService.getDukanByOwnerId(userId)
         return ResponseEntity.ok(DukanStatuesResponse(dukan.name, dukan.status))
     }
 }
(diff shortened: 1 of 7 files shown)


haideraljawasim (approved):

ElsayedMagdy122 (approved):

mostafa-n3ma (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
```diff
@@ -22,4 +25,12 @@ class DukanController(
         val available = dukanService.isDukanNameAvailable(name)
         return ResponseEntity.ok(DukanNameAvailabilityResponse(available))
     }
+
+    @GetMapping("/statues")
+    fun getDukanStatues(
+        @AuthenticationPrincipal ownerId: UUID,
```
devfalah: This is `userId` but we save it as `ownerId`, I suggest renaming it to `userId`

devfalah (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
Karrar-Mohammed: since this is the principal so i agree on renaming it to userId

Karrar-Mohammed: because at this point of the process it's the id of the user that making the request, but we can name it ownerId in the database

Karrar-Mohammed (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
TarekIdrees: Noted

PR #24, opened 2025-09-17
IDN-26 implement the refresh token endpoint

AimanYosofiUlt:


identity/src/main/kotlin/identity/api/controller/IdentityController.kt
```diff
@@ -1,20 +1,33 @@
...
 @RestController
 @RequestMapping("/identity")
 class IdentityController(
     val authenticationService: AuthenticationService
 ) {
+    @ExceptionHandler(BadCredentialsException::class)
+    fun handleBadCredentialsException(): ResponseEntity<String> {
+        return ResponseEntity.status(HttpStatus.UNAUTHORIZED).body("Invalid phone number or password")
+    }
+
```
Karrar-Mohammed, line 22: for now keep this function here but we will merge the exception handler PR soon, so after we merge the PR you can see how we are handling the exceptions in other module and do the same

identity/src/main/kotlin/identity/api/dto/RefreshTokenRequest.kt
```diff
@@ -0,0 +1,8 @@
+package identity.api.dto
```
Karrar-Mohammed, line 1: you need fix the packages in the identity module, add the net/thechance to the root of the module so it will be like that

net/thechance/identity

check the dukan module for more details

Karrar-Mohammed (approved):

nabilBouzineDev (approved):
LGTM 👍

poula22 (approved):

PR #25, opened 2025-09-17
Feature/TRN-18 user categories

mohamedd-hassan:
created a `TrendUser` table to save user info related to trend feature
created a `TrendUserRepository` and `TrendUserService` and `CategoryService`
added a controller for posting user interests


trends/src/main/kotlin/net/thechance/trends/api/controller/CategoryController.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.http.ResponseEntity
+import org.springframework.security.core.annotation.AuthenticationPrincipal
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.util.UUID
+
+@RestController
+@RequestMapping("/trends")
```
iZyadNasser: I think you can put the category path here so ("/trends/category") since this controller is for category end points only anyway

trends/src/main/kotlin/net/thechance/trends/api/controller/CategoryController.kt
```diff
@@ -0,0 +1,39 @@
...
+class CategoryController(
+    private val categoryService: CategoryService,
+    private val trendUserService: TrendUserService,
+) {
+
+    @PostMapping("/interests")
```
iZyadNasser: let's stick to the "category" name in all our code so we don't confuse ourselves with category/interest

iZyadNasser: Also, is it better to use plural form in endpoints (categories-interests), or single form (category, interest)? I honestly don't know what is better but we need to choose one of them to unify our naming convention across all the feature

trends/src/main/kotlin/net/thechance/trends/api/controller/CategoryController.kt
```diff
@@ -0,0 +1,39 @@
...
+    fun postUserInterests(
```
iZyadNasser: Again, let's stick with categories everywhere

trends/src/main/kotlin/net/thechance/trends/entity/TrendUser.kt
```diff
@@ -0,0 +1,29 @@
...
+import jakarta.persistence.JoinColumn
+import jakarta.persistence.JoinTable
+import jakarta.persistence.ManyToMany
+import jakarta.persistence.Table
+import java.util.Collections.emptySet
+import java.util.UUID
+
+@Table(name = "user", schema = "trends")
+@Entity
+data class TrendUser(
```
ahmeddsamii, line 16: why the name of this is TrendUser, and you're saving categories here? and i notice the name of table is user

trends/src/main/kotlin/net/thechance/trends/service/TrendUserService.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import net.thechance.trends.entity.Category
+import net.thechance.trends.repository.TrendUserRepository
+import org.springframework.stereotype.Service
+import org.springframework.transaction.annotation.Transactional
+import java.util.UUID
+
+@Service
+@Transactional
+class TrendUserService(
```
ahmeddsamii, line 17: same here, the name is TrendUserService, and you're saving categories, so you could simply move it to CategoryService

trends/src/main/kotlin/net/thechance/trends/entity/TrendUser.kt
mohamedd-hassan: the reason is because when running the server spring boot has a problem managing the bean because of duplicated naming
here this is the user info for trend feature and I'm saving the trends the user chose here

ahmeddsamii: okay but you are managing categories not trends

trends/src/main/kotlin/net/thechance/trends/service/TrendUserService.kt
mohamedd-hassan: the reason is this TrendUser will be expanded on in the future to include name and profile and picture and more information so this is the service responsible for saving the User chosen categories

trends/src/main/kotlin/net/thechance/trends/entity/TrendUser.kt
mohamedd-hassan: the entire feature is named trend and we have a user that we need to save their selected categories
i cant save them in the categories table because that is user information (user has a name, an id, a picture, some interests, favourite trends, etc) so this is the initial trend user table for our feature which for now will just contain his chose interests

iZyadNasser (approved):

trends/src/main/kotlin/net/thechance/trends/api/controller/category/CategoryController.kt
```diff
@@ -0,0 +1,39 @@
...
+
+@RestController
+@RequestMapping("/trends/categories")
+class CategoryController(
+    private val categoryService: CategoryService,
+    private val trendUserService: TrendUserService,
+) {
+
+    @PostMapping
+    fun postUserCategories(
```
Karrar-Mohammed: name the function with a clear action not by the http method, like addUserCategory createUserCategory or something like that

trends/src/main/kotlin/net/thechance/trends/entity/Category.kt
```diff
@@ -13,9 +13,9 @@ data class Category(
     @Column(columnDefinition = "uuid", nullable = false, updatable = false)
     val id: UUID = UUID.randomUUID(),
 
-    @Column(name = "name", nullable = false, unique = true)
+    @Column(nullable = false, unique = true)
     val name: String,
 
-    @Column(name = "emoji", nullable = false, unique = false)
+    @Column(nullable = false, unique = false)
```
Karrar-Mohammed: if the unique is false then i think you can remove it because this is the default value

trends/src/main/kotlin/net/thechance/trends/api/controller/category/CategoryController.kt
mohamedd-hassan: oki done

trends/src/main/kotlin/net/thechance/trends/entity/Category.kt
mohamedd-hassan: done

trends/src/main/kotlin/net/thechance/trends/entity/TrendUser.kt
Karrar-Mohammed, line 16: i think it's better to call it TrendUserCategory, because this is related to the category not all the user details

mohamedd-hassan: in the future we will have other user info like name and profile picture in future stories not just categories

trends/src/main/kotlin/net/thechance/trends/entity/TrendUser.kt
```diff
@@ -0,0 +1,29 @@
...
+    val userId: UUID,
+
+    @ManyToMany(fetch = FetchType.EAGER)
+    @JoinTable(
+        name = "user_categories",
+        joinColumns = [JoinColumn(name = "user_id")],
+        inverseJoinColumns = [JoinColumn(name = "category_id")],
+        schema = "trends",
+    )
+    val categories: MutableSet<Category> = emptySet(),
```
Karrar-Mohammed, line 28: how many tables will be created here ? because we only need a table to save the user Id with category Id but here you are creating entity then using the @ManyToMany annotation which will create another table as far as i know, so i think we can do it by only one table which will be enough

mohamedd-hassan: we will have other user info here like we discussed in the meeting like the name and profile picture and maybe favourited reels etc

trends/src/main/kotlin/net/thechance/trends/service/TrendUserService.kt
```diff
@@ -0,0 +1,24 @@
...
+    private val trendUserRepository: TrendUserRepository
+) {
+    fun saveCategoriesToUser(userId: UUID, categories: List<Category>) {
+        val trendUser = trendUserRepository.findById(userId).orElseThrow{ TrendUserNotFoundException() }
```
Karrar-Mohammed: i don't think this is correct because at the beginning will be there is no record in the table, so i'm not sure if this will be working as expected

trends/src/main/kotlin/net/thechance/trends/service/TrendUserService.kt
```diff
@@ -0,0 +1,24 @@
...
+        val existingCategoryIds = trendUser.categories.map { it.id }.toSet()
+        val newCategories = categories.filterNot { it.id in existingCategoryIds }
+
+        trendUser.categories.clear()
+        trendUser.categories.addAll(newCategories)
+        trendUserRepository.save(trendUser)
```
Karrar-Mohammed: i think all we need is to save the category with user id so i think it can be simpler

trends/src/main/kotlin/net/thechance/trends/service/TrendUserService.kt
mohamedd-hassan: yes this is true i think i should maybe insert if not found instead of throwing an exception right? the thing is that I had in mind that i will save this info from the identity event that will get sent so should I just for now save here?

trends/src/main/kotlin/net/thechance/trends/service/TrendUserService.kt
mohamedd-hassan: yes I am just clearing any saved categories he had earlier to act as new addition instead of update

trends/src/main/kotlin/net/thechance/trends/service/TrendUserService.kt
Karrar-Mohammed: we don't have such an event write now and i don't we will have it soon, so this will not work, so let's fix it

trends/src/main/kotlin/net/thechance/trends/api/dto/category/SubmitUserCategoriesResponse.kt
```diff
@@ -0,0 +1,9 @@
+package net.thechance.trends.api.dto.category
+
+import java.util.UUID
+
+data class SubmitUserCategoriesResponse(
+    val userId: UUID,
+    val categoryIds: List<UUID>,
+    val message: String
```
Karrar-Mohammed: why do we need to return userId and message ? how this will help the mobile apps ?

trends/src/main/kotlin/net/thechance/trends/api/controller/user/TrendUserController.kt
```diff
@@ -0,0 +1,27 @@
...
+) {
+
+    @GetMapping("/user/categories/status")
+    fun getDoesUserHaveCategories(
+        @AuthenticationPrincipal userId: UUID,
+    ): ResponseEntity<DoesUserHaveCategoriesResponse> {
+        val doesUserHaveCategories = trendUserService.getDoesUserHaveCategories(userId)
+        val response = DoesUserHaveCategoriesResponse(value = doesUserHaveCategories)
+        return ResponseEntity.ok(response)
+    }
```
Karrar-Mohammed, line 26: why do we need this end point ?

trends/src/main/kotlin/net/thechance/trends/service/CategoryService.kt
```diff
@@ -1,13 +1,30 @@
...
+        if (categoryIds.isEmpty()) {
+            throw InvalidTrendInputException()
+        }
+
+        val categories = categoryRepository.findAllById(categoryIds) ?: throw TrendCategoryNotFoundException()
+        if(categories.size != categoryIds.size) {
+            throw TrendCategoryNotFoundException()
+        }
+        return categories
+    }
```
Karrar-Mohammed: if this function it's idea to only get the category so we can save a new record in the user table, then there is something called proxy that you can create a proxy of the category with the category id only and then save the record in the user table without getting the categories list, so i think we don't need this function

trends/src/main/kotlin/net/thechance/trends/api/dto/category/SubmitUserCategoriesResponse.kt
mohamedd-hassan: I thought that it might be useful but thinking about it now I can just remove this extra information

done

trends/src/main/kotlin/net/thechance/trends/api/controller/user/TrendUserController.kt
mohamedd-hassan: In mobile app when first opening the trend feature, if the user has any categories saved we display the trends screen and if it's their first time we give them the choose category screen

so we need a way to know if the user has any saved categories or not

trends/src/main/kotlin/net/thechance/trends/service/CategoryService.kt
mohamedd-hassan: thank you i didnt know about proxies but I just applied them here and deleted the function like you said

trends/src/main/kotlin/net/thechance/trends/api/dto/trendUser/DoesUserHaveCategoriesResponse.kt
```diff
@@ -0,0 +1,5 @@
+package net.thechance.trends.api.dto.trendUser
+
+data class DoesUserHaveCategoriesResponse(
+    val value: Boolean
```
Karrar-Mohammed: i suggest to name this parameter as "hasCategory", it might look unneeded because it can be easily understood what is it from the class name, but in the response will not be easy to understand what this value, so i suggest to rename just for the name in the json response

Karrar-Mohammed (approved):

trends/src/main/kotlin/net/thechance/trends/api/dto/trendUser/DoesUserHaveCategoriesResponse.kt
mohamedd-hassan: okay done

ahmeddsamii (approved):

PR #26, opened 2025-09-17
Feature/ Uplaod Dukan Image 

haideraljawasim:
# What have been done :
- implement endpoint to upload dukan image 

`/dukan/image`

### DIFF
--- a/dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
@@ -22,7 +22,7 @@ data class Dukan(
     )
     val categories: Set<DukanCategory>,
     @Column(name = "image_url", nullable = true)
-    var imageUrl: String? = null,
+    val imageUrl: String? = null,
     @Column(name = "address", nullable = false)
     val address: String,
     @Column(name = "latitude", nullable = false)
--- a/dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
@@ -22,7 +22,7 @@ data class Dukan(
     )
     val categories: Set<DukanCategory>,
     @Column(name = "image_url", nullable = true)
-    val imageUrl: String? = null,
+    var imageUrl: String? = null,
     @Column(name = "address", nullable = false)
     val address: String,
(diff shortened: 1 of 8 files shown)


dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
```diff
@@ -22,7 +22,7 @@ data class Dukan(
     )
     val categories: Set<DukanCategory>,
     @Column(name = "image_url", nullable = true)
-    val imageUrl: String? = null,
+    var imageUrl: String? = null,
```
devfalah: Please keep the table columns immutable

```suggestion
    val imageUrl: String? = null,
```

dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
```diff
@@ -47,6 +50,15 @@ class DukanService(
         )
     }
 
+    @Transactional
+    fun uploadDukanImage(ownerId: UUID, file: MultipartFile): String {
+        val dukan = dukanRepository.findByOwnerId(ownerId)?: throw DukanNotFoundException()
+        val imageUrl = file.originalFilename.orEmpty() // TODO (Upload the image to real storage service)
+        dukan.imageUrl = imageUrl
+
+        return imageUrl
```
devfalah:
```suggestion
        return  dukanRepository.save(dukan.copy(imageUrl= imageUrl))
```

haideraljawasim: consider it done, but can i ask why we here return the whole dukan object why not only image url

dukan/src/main/kotlin/net/thechance/dukan/entity/Dukan.kt
haideraljawasim: done

dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
devfalah: Aha you are right, we don’t need to return the whole object, my point was about saving the updated dukan

devfalah (approved):

TarekIdrees (approved):

Karrar-Mohammed (approved):

PR #27, opened 2025-09-18
Feature/IDN-25 ensure prevention of brute force attacks

poula22:
Add new entity login log to track user attempts to login. Then verify if the same `Ip address` tries to make login many times to block him. Implementation includes:
- add `LoginLog` entity
- add `LoginLogRepository`
- add `LoginLogService`
- use `LoginLogService` in `AuthenticationService` to verify if user is blocked or not and add login attempts logs


identity/src/main/kotlin/identity/api/controller/IdentityController.kt
```diff
@@ -0,0 +1,25 @@
...
+@RequestMapping("/identity")
+class IdentityController(
+    private val authenticationService: AuthenticationService
+) {
+    @PostMapping("/login")
+    fun login(
+        @RequestBody @Valid request: AuthRequest,
+        httpRequest: HttpServletRequest
+    ): ResponseEntity<AuthResponse> {
+        val ipAddress = httpRequest.remoteAddr ?: throw IllegalStateException("Invalid IP")
```
AimanYosofiUlt, line 29: I think you can create a custom exception instead of throwing `IllegalStateException`

poula22: I did that in another branch for error handling

identity/src/main/kotlin/identity/entity/LoginLog.kt
```diff
@@ -0,0 +1,29 @@
...
+import java.time.Instant
+import java.util.*
+
+@Entity
+@Table(name = "login_log", schema = "identity")
+class LoginLog(
+    @Id
+    @GeneratedValue(strategy = GenerationType.UUID)
+    @Column(name = "id", nullable = false)
+    val id: UUID? = null,
```
Karrar-Mohammed: why the default value of the id is null ? i think you can make it as UUID.randomUUID()

Karrar-Mohammed: it's better to not make the id as nullable

identity/src/main/kotlin/identity/service/AuthenticationService.kt
```diff
@@ -1,32 +1,69 @@
...
+        val loginLog = LoginLog(user = user, isSuccess = isSuccess, ipAddress = ipAddress)
+        loginLogService.addLoginLog(loginLog)
+    }
+
+    private fun isUserBlocked(ipAddress: String): Boolean {
+        val loginLogs = loginLogService.getLoginLogsByIpAddress(ipAddress, 5)
+            .filter { !it.isSuccess }
+            .ifEmpty { return false }
+
+        val lastTimeToLogin = loginLogs.first().loginTime
```
Karrar-Mohammed, line 65: how can you be sure that the first item in the list is the last time login ? i suggest you should add sorting so you can be sure about that

identity/src/main/kotlin/identity/service/AuthenticationService.kt
```diff
@@ -1,32 +1,69 @@
...
+        val now = Instant.now()
+        val durationSinceLastLogin = Duration.between(lastTimeToLogin, now)
+        val firstTimeToLogin = loginLogs.last().loginTime
+        val durationBetweenFirstAndLastLogin = Duration.between(firstTimeToLogin, lastTimeToLogin)
+        if (loginLogs.size < MAX_LOGIN_ATTEMPTS) return false
+        if (durationSinceLastLogin.toMinutes() >= MAX_BLOCK_TIME_IN_MINUTES) return false
+        return durationBetweenFirstAndLastLogin.toMinutes() <= BLOCK_TIME_IN_MINUTES
```
Karrar-Mohammed, line 75: can we separate this logic into multiple small functions ?

identity/src/main/kotlin/identity/service/AuthenticationService.kt
```diff
@@ -1,32 +1,69 @@
...
+    private fun addUserToLogs(
+        user: User,
+        isSuccess: Boolean,
+        ipAddress: String
+    ) {
+        val loginLog = LoginLog(user = user, isSuccess = isSuccess, ipAddress = ipAddress)
+        loginLogService.addLoginLog(loginLog)
+    }
+
+    private fun isUserBlocked(ipAddress: String): Boolean {
```
Karrar-Mohammed, line 52: if you can try to write unit test for this logic, i think it's worth to write tests

identity/src/main/kotlin/identity/service/AuthenticationService.kt
poula22: I am already doing a sort by loginDate but in loginLogService. Should I make it more resuable?

identity/src/main/kotlin/identity/service/AuthenticationService.kt
poula22: I will do that, but in exception handling branch to test them both

identity/src/main/kotlin/identity/entity/LoginLog.kt
poula22: I want to make database responsible for creating new ids, but I don't know the right approach here.

This will happen by the annotation generatedValue and nullable false

identity/src/main/kotlin/identity/service/AuthenticationService.kt
poula22: Great I will do that

identity/src/main/kotlin/identity/service/AuthenticationService.kt
poula22: I add these 2 parameters to getLoginLogsByIpAddress function:
sortedBy: String = LoginLog::loginTime.name,
sortedDirection: Sort.Direction = Sort.Direction.DESC

identity/src/main/kotlin/identity/entity/LoginLog.kt
poula22: I changed it

identity/src/main/kotlin/identity/service/AuthenticationService.kt
poula22: done

identity/src/main/kotlin/identity/entity/LoginLog.kt
Karrar-Mohammed: we are not making the database responsible for generating the id, currently our backend is generating the ids

Karrar-Mohammed (approved):

AhmadKharfan (approved):

Ahmed7osny1 (approved):

PR #29, opened 2025-09-18
- add the BalanceController & the WalletService

EL-MANCY:


Al-Qassim (approved):

wallet/src/main/kotlin/net/thechance/wallet/api/dto/BalanceDto.kt
Malak187, line 1: I think you can call it BalanceResponse for better readability

Malak187 (approved):
Great work Abdelrahman 👏

PR #36, opened 2025-09-19
CHT-34-rename user related entities and update references and fix a bug

Joseph-Sameh-0:


chat/src/main/kotlin/net/thechance/chat/entity/ContactUser.kt
```diff
@@ -7,8 +7,8 @@ import jakarta.persistence.Table
 import java.util.UUID
 
 @Entity
-@Table(name = "users", schema = "chat")
-data class User(
+@Table(name = "contactUser", schema = "chat")
```
devfalah: The table name `contactUser` doesn’t follow our tables naming conventions you can rename it to something like: `user_contacts`

chat/src/main/kotlin/net/thechance/chat/service/ContactUserService.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import net.thechance.chat.entity.ContactUser
+import net.thechance.chat.repository.ContactUserRepository
+import org.springframework.stereotype.Service
+
+@Service
+class ContactUserService(
+    private val contactUserRepository: ContactUserRepository
+) {
+    init { // Initialize with some contact users if the repository is empty for testing purposes
```
devfalah: Please dont push testing code to production. once the app is stable on the development server we can add dummy data to the server database

chat/src/main/kotlin/net/thechance/chat/entity/ContactUser.kt
Joseph-Sameh-0: in the database the name will be user_contacts
image

devfalah: we are naming our tables in plural so contact_user does not follow that

chat/src/main/kotlin/net/thechance/chat/entity/ContactUser.kt
```diff
@@ -7,8 +7,8 @@ import jakarta.persistence.Table
 import java.util.UUID
 
 @Entity
-@Table(name = "users", schema = "chat")
-data class User(
+@Table(name = "users_of_contacts", schema = "chat")
```
devfalah, line 10: I think you are saving user contacts in this table so the better name is user_contacts instead of users_of_contacts if someone read this name without context he will think its a join table between users and contacts

chat/src/main/kotlin/net/thechance/chat/entity/ContactUser.kt
```diff
@@ -7,8 +7,8 @@ import jakarta.persistence.Table
...
+data class ContactUser(
```
devfalah, line 11: It would be better to rename to UserContact instead of ContactUser

Joseph-Sameh-0: it doesn't save contact but user of the contact

chat/src/main/kotlin/net/thechance/chat/entity/ContactUser.kt
Joseph-Sameh-0: no I save the user not the contact
I really use it to join with it when I get the contacts

devfalah (approved):

omer1998 (approved):

AhmadKharfan (approved):

PR #43, opened 2025-09-20
Feature/IDN-28 implement identity exception handler

poula22:
handle Authentication Exceptions
add database `liquibase` change log
add `unit test` to `login` logic


AimanYosofiUlt (approved):
Well done

identity/src/main/kotlin/identity/api/controller/IdentityController.kt
```diff
@@ -35,4 +34,48 @@ class IdentityController(
...
+                ResponseEntity
+                    .status(HttpStatus.UNAUTHORIZED)
+                    .body(ErrorResponse("Invalid refresh token"))
+            }
+
+            else -> {
+                ResponseEntity
+                    .status(HttpStatus.INTERNAL_SERVER_ERROR)
+                    .body(ErrorResponse("unknown error happened"))
+            }
```
Karrar-Mohammed: i suggest to create a separate function to handle each exception, it will easier to understand and easier to add new exception

identity/src/main/kotlin/identity/api/controller/IdentityController.kt
```diff
@@ -35,4 +34,48 @@ class IdentityController(
...
+        }
+    }
+
+    @ExceptionHandler(Exception::class)
+    fun handleException(exception: Exception): ResponseEntity<ErrorResponse?> {
+        logger.error("Unknown error happened: ${exception.message}", exception)
+        return ResponseEntity
+            .status(HttpStatus.INTERNAL_SERVER_ERROR)
+            .body(ErrorResponse("unknown error happened"))
+    }
```
Karrar-Mohammed: i suggest to move all these function for handling exception into a controller advice class that responsible for handling these exceptions, so we don't make the controller messy, let's the controller class handle the request and the response, and the controller advice for handling the exceptions

identity/src/main/resources/db-identity/changelog/changelogs/changelogs_001.yaml
```diff
@@ -25,4 +25,22 @@ databaseChangeLog:
...
+      id: 3-identity
+      author: poula
+      comment: Create login_log table for identity
+      changes:
+        - sql:
+            create table identity.login_log
+            (
+            id         uuid                    not null
+            constraint pk_login_log
+            primary key,
```
Karrar-Mohammed: why don't you putt all of these in one line, i think it will be easier to read

Karrar-Mohammed (commented):
great work, let's just fix these simple comments before merging the PR

identity/src/main/kotlin/identity/api/controller/IdentityController.kt
poula22: Done

identity/src/main/resources/db-identity/changelog/changelogs/changelogs_001.yaml
poula22: Done

identity/src/main/kotlin/identity/api/controller/IdentityController.kt
poula22: Done

Karrar-Mohammed (approved):

AhmadKharfan (approved):

PR #48, opened 2025-09-22
Feature / Dukan shelf creation endpoint 

haideraljawasim:
# What have been done :
- Implement dukan shelf creation endpoint 
- endpoint 
`/dukan/shelf/create`

Request body 
```
{
"title": String 
}
```

### DIFF
--- a/dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanShelfController.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanShelfController.kt
@@ -1,0 +1,33 @@
+package net.thechance.dukan.api.controller
+
+import jakarta.validation.Valid
+import net.thechance.dukan.api.dto.DukanShelfCreationRequest
+import net.thechance.dukan.entity.DukanShelf
+import net.thechance.dukan.service.DukanShelfService
+import org.springframework.http.ResponseEntity
+import org.springframework.security.core.annotation.AuthenticationPrincipal
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.security.Principal
+import java.util.UUID
+
+@RestController
+@RequestMapping("/dukan/shelf")
+class DukanShelfController(
+    val dukanShelfService: DukanShelfService
+) {
+
+    @PostMapping("/create")
+    fun createShelf(
+        @Valid @RequestBody requestBody: DukanShelfCreationRequest,
+        @AuthenticationPrincipal ownerId: UUID,
+    ): ResponseEntity<DukanShelf> {
+        val shelf = dukanShelfService.createShelf(
+            ownerId = ownerId,
+            title = requestBody.title,
+        )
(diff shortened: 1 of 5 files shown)


ElsayedMagdy122 (approved):

ElsayedMagdy122:
Great Work bro 👏

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanShelfController.kt
```diff
@@ -0,0 +1,31 @@
...
+    @PostMapping("/create")
+    fun createShelf(
+        @Valid @RequestBody requestBody: DukanShelfCreationRequest,
+        @AuthenticationPrincipal ownerId: UUID,
+    ): ResponseEntity<Unit> {
+        dukanShelfService.createShelf(
+            ownerId = ownerId,
+            title = requestBody.title,
+        )
+        return ResponseEntity.ok().build()
```
TarekIdrees, line 29: For consistency, as we did with createDukan, we might also need to return the Shelf object in the response body along with the OK status. Does that make sense to you?

haideraljawasim: To be honest I implemented it at first but then i noticed when fetching a `Shelf` it will also return the `Dukan` which we don't  really need , so i just change it to return OK status instead , since returning a the `shelf` with the `Dukan` will just be unnecessary overhead , also considered creating a response body to return just the shelf but that didn't seem necessary either, but np if you see returning the `Shelf` might be useful i can add it to the response body

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanShelfController.kt
```diff
@@ -0,0 +1,31 @@
...
+@RestController
+@RequestMapping("/dukan/shelf")
+class DukanShelfController(
+    val dukanShelfService: DukanShelfService
+) {
+
+    @PostMapping("/create")
+    fun createShelf(
+        @Valid @RequestBody requestBody: DukanShelfCreationRequest,
+        @AuthenticationPrincipal ownerId: UUID,
```
devfalah: please rename it to userId

devfalah: we save it as ownerId but from AuthenticationPrincipal we get userId

devfalah (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanShelfController.kt
haideraljawasim: done

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanShelfController.kt
TarekIdrees: I had a small discussion with Falah  about this point in order to arrange the excepted body as a return and we decided to keep the existing eindpoints as it is but then later on we need to make a refinement to accomplished the point you mentioned and prevent unnecessary returned objects.

TarekIdrees (approved):

PR #49, opened 2025-09-24
Feature/cht 38 prepare entities and repository

Joseph-Sameh-0:
## What is the PR scope?
This PR adds the core data model for a chat feature. It includes new tables (Chat, Message), repositories

## Why do we need that?
To allowing users to send messages in one-on-one conversations, track who has read what, and link contacts to actual user accounts


monaayman123 (approved):

omer1998 (approved):
good job Joseph, it seems good to me

ahmedfikry24 (approved):

raouf-kamel (approved):

chat/src/main/kotlin/net/thechance/chat/entity/Contact.kt
```diff
@@ -12,7 +12,7 @@ import java.util.UUID
     name = "contacts", schema = "chat",
     uniqueConstraints = [UniqueConstraint(columnNames = ["contact_owner_id", "phone_number"])]
 )
-data class Contact(
+open class Contact(
```
devfalah: why did u make it open class?

chat/src/main/kotlin/net/thechance/chat/entity/GroupConversation.kt
```diff
@@ -0,0 +1,23 @@
...
+    @JoinColumn(
+        name = "chat_id",
+        referencedColumnName = "id",
+        nullable = false,
+        foreignKey = ForeignKey(name = "fk_group_chat_chat")
+    )
+    val chat: Chat,
+
+    @Column(name = "group_name", nullable = true)
+    val groupName: String?,
```
devfalah: is it possible for users to create group without name?

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
```diff
@@ -0,0 +1,77 @@
...
+    )
+    FROM Chat c
+    LEFT JOIN GroupChat gc ON c.id = gc.chat.id
+    LEFT JOIN ChatParticipants p ON c.id = p.id.chatId AND c.isGroup = false
+    LEFT JOIN ContactUser u ON p.id.userId = u.id AND c.isGroup = false
+    LEFT JOIN Contact ct ON ct.phoneNumber = u.phoneNumber AND ct.contactOwnerId = :userId AND c.isGroup = false
+    WHERE c.id = :conversionId AND (c.isGroup = true OR p.id.userId <> :userId)
+    """
+    )
+    fun getChatById(conversionId: UUID, userId: UUID): ChatModel?
```
Karrar-Mohammed: why do we need to write the query by ourself ? i think we can use the ORM to handle the join for us

Karrar-Mohammed: i don't recommend to keep writing the queries like that

chat/src/main/kotlin/net/thechance/chat/entity/Contact.kt
Joseph-Sameh-0: The IDE makes warning when I make it data class and it recommends to be open class

chat/src/main/kotlin/net/thechance/chat/entity/GroupConversation.kt
Joseph-Sameh-0: You are right I will fix it

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
Joseph-Sameh-0: No the ORM can't, I made the database normalized and the ORM can't handle the multiple joins or the complex queries like this

chat/src/main/kotlin/net/thechance/chat/api/dto/ContactResponse.kt
```diff
@@ -8,5 +8,6 @@ data class ContactResponse(
     val lastName: String,
     val phoneNumber: String,
     val isMenaUser: Boolean,
+    val userId: UUID?,
```
Karrar-Mohammed: why do we need to return the userId ? and why it's nullable ?

chat/src/main/kotlin/net/thechance/chat/entity/Contact.kt
Karrar-Mohammed: no need to make it open by ourself there are plugins do that for us so keep it as data class

chat/src/main/kotlin/net/thechance/chat/api/dto/ContactResponse.kt
Joseph-Sameh-0: the contact can be mena user so the client will send the user Id he want to chat with 
it will be null if the contact not mena user

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
Karrar-Mohammed: i think it can do that, working like this way is not type safe and is not scalable, imagine we add new field to the ChatModel how you will handle that or the other developer how he will know we are doing the query in that way, so i don't recommend this approach because of that

chat/src/main/kotlin/net/thechance/chat/entity/GroupChat.kt
```diff
@@ -0,0 +1,23 @@
...
+
+    @Id
+    @OneToOne
+    @JoinColumn(
+        name = "chat_id",
+        referencedColumnName = "id",
+        nullable = false,
+        foreignKey = ForeignKey(name = "fk_group_chat_chat")
+    )
+    val chat: Chat,
```
Karrar-Mohammed: this field type is Chat why are you annotating it as @Id how it can be primary key ? also here you are using the join annotation then why you don't use it with the other entities ?

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
Joseph-Sameh-0: to make the ORM work with it I should make it like a relation, and I don't want to make relation with user table because the user in user table can be deleted, I  don't want to delete every thing the user has a relation with if it's row deleted

what do you think ?

chat/src/main/kotlin/net/thechance/chat/entity/GroupChat.kt
Joseph-Sameh-0: it is an id because this is a `weak entity` related only with one chat, it's a PK and FK in the same time
I use it in the join query 
image

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
Joseph-Sameh-0: The ORM has some limitations to do this, you can try

Joseph-Sameh-0: I will try to change it now

chat/src/main/kotlin/net/thechance/chat/service/MessagingService.kt
```diff
@@ -0,0 +1,38 @@
...
+            if (messages.isNotEmpty()) {
+                messageRepository.saveAll(messages.onEach { it.readByUsers += user })
+            }
+            page++
+        } while (messages.size == pageSize)
+    }
+
+    companion object {
+        const val PAGE_SIZE = 500
+    }
```
mosayed01: will be better if we make it in a single db query

Joseph-Sameh-0: it will be native query to do insert
is it better ?

Karrar-Mohammed: why do we need to mark all the messages of a chat at once as read ? shouldn't be by the message ?

Joseph-Sameh-0: it's not all messages the unread messages when open the chat 
@Karrar-Mohammed

Karrar-Mohammed: but it doesn't make sense image the unread messages more than 100 and i only read 10 messages so why do we mark all the 100 messages as read ? the reasonable is to mark only the 10 messages that i read

Joseph-Sameh-0: we don't have requirement like this 
for now we consider the all the chat is read if the user opens it 
@Karrar-Mohammed

Joseph-Sameh-0: we can plan to make it like this in a coming sprint

Joseph-Sameh-0: And the other user should know that i read the message so if it like this we will send event in the web socket for every message the user read 

we simplified this and will send one event tells that the user read and mark all as reed

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
```diff
@@ -0,0 +1,11 @@
+package net.thechance.chat.repository
+
+import net.thechance.chat.entity.Chat
+import net.thechance.chat.entity.ContactUser
+import org.springframework.data.jpa.repository.JpaRepository
+import java.util.UUID
+
+interface ChatRepository : JpaRepository<Chat, UUID> {
+    fun findByIdIs(id: UUID): Chat?
```
Karrar-Mohammed: this can be removed

Karrar-Mohammed (approved):
good job 👏

PR #53, opened 2025-09-25
Feature/WLT-25 transactions filter endpoint

Malak187:
## **PR Scope**

- Created a transaction history endpoint with support for filtering by status, type and dates.
- Created end point to get the user's first transaction date.

## Why do we need that?

- To allow users to view their transaction history in a structured and paginated way.
- To provide filtering options (by type/status/dates) for easier tracking of specific transactions.
- To return the earliest transaction date, which can be useful for using it as default date in start date picker.

## End points with the request and response body:

- Example of response when fetching all user's transactions:

### `GET /wallet/transactions?page=0&size=4`

```json
{
    "totalElements": 8,
    "page": 0,
    "pageSize": 4,
    "totalPages": 2,
    "transactions": [
         {
            "id": "0f0054b2-624b-45ca-9060-ee1d9939613e",
            "senderName": "hager",
            "receiverName": "malak",
            "status": "SUCCESS",
            "type": "SENT",
            "createdAt": "2025-09-22T22:50:36.590322",
            "amount": 473.0
        },
        {
            "id": "53e695d0-f961-465f-bee7-b2e022a985e2",
            "senderName": "malak",
            "receiverName": "hager",
            "status": "SUCCESS",
            "type": "RECEIVED",
            "createdAt": "2025-09-22T22:48:36.380795",
            "amount": 595.0
        },
        {
            "id": "cfcb62ba-f9b1-485b-9ec8-94864286f40c",
            "senderName": "hager",
            "receiverName": "malak",
            "status": "FAILED",
            "type": "SENT",
            "createdAt": "2025-09-22T22:49:06.370236",
            "amount": 49.0
        },
        {
            "id": "a65099fa-6f8f-4fc8-9db7-482bf6b7de9d",
            "senderName": "malak",
            "receiverName": "hager",
            "status": "FAILED",
            "type": "RECEIVED",
            "createdAt": "2025-09-22T22:47:43.709066",
            "amount": 120.0
        }
    ]
}
```

- Example of response when fetching **filtered** user's transactions:

### `GET /wallet/transactions?type=ONLINE_PURCHASE&status=SUCCESS`

```json
{
    "totalElements": 1,
    "page": 0,
    "pageSize": 20,
    "totalPages": 1,
    "transactions": [
        {
            "id": "37d5e296-9fa4-4cab-b346-3c966e2de696",
            "senderName": "hager",
            "receiverName": "Adidas",
            "status": "SUCCESS",
            "type": "ONLINE_PURCHASE",
            "createdAt": "2025-09-23T04:16:53.621524",
            "amount": 6.0
        }
    ]
}
```

- Example of response when getting the date of user's first transaction

### `GET /wallet/transactions/first-date`

```json
{
    "firstTransactionDate": "2025-09-22"
}
```


wallet/src/main/resources/db-wallet/changelog/changelogs/changelogs_001.yaml
```diff
@@ -50,4 +50,56 @@ databaseChangeLog:
...
+      author: malakraef
+      comment: Create wallet_user table to store replicated user and dukan info
+      changes:
+        - sql:
+            sql: |
+              CREATE TABLE wallet.wallet_user (
+                user_id UUID PRIMARY KEY,
+                user_name VARCHAR NOT NULL,
+                dukan_name VARCHAR
+              );
```
Al-Qassim: add these to a new file and call it 
changelogs_002.yaml

wallet/src/main/resources/db-wallet/changelog/changelogs/changelogs_001.yaml
```diff
@@ -50,4 +50,56 @@ databaseChangeLog:
...
+                  constraints:
+                    nullable: false
+              - column:
+                  name: status
+                  type: VARCHAR(50)
+                  constraints:
+                    nullable: false
+
+  - changeSet:
+      id: 5-remove-type-column
```
Al-Qassim: make the id = 5-wallet to be consistent

wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/TransactionResponse.kt
```diff
@@ -0,0 +1,23 @@
...
+import net.thechance.wallet.entity.Transaction
+import org.springframework.data.domain.Page
+import java.time.LocalDateTime
+import java.util.*
+
+data class TransactionResponse(
+    val id: UUID,
+    val senderId: UUID,
+    val senderName: String,
+    val receiverId: UUID,
```
Karrar-Mohammed: why do we need to return the senderId and the receiverId ?

wallet/src/main/kotlin/net/thechance/wallet/repository/TransactionRepository.kt
```diff
@@ -1,15 +1,71 @@
...
+              AND (t.senderId = :currentUserId 
+                   OR t.receiverId = :currentUserId)
+              AND (
+                :type IS NULL
+                OR (:type = 'SENT' AND t.senderId = :currentUserId AND receiver.dukanName IS NULL)
+                OR (:type = 'RECEIVED' AND t.receiverId = :currentUserId AND receiver.dukanName IS NULL)
+                OR (:type = 'ONLINE_PURCHASE' AND receiver.dukanName IS NOT NULL)
+              )
+        """
+    )
```
Karrar-Mohammed, line 32: why do we need to write the query by ourself and not let the orm handle it for us ?

wallet/src/main/kotlin/net/thechance/wallet/repository/transaction/TransactionProjection.kt
```diff
@@ -0,0 +1,19 @@
...
+    val id: UUID,
+    val createdAt: LocalDateTime,
+    val amount: BigDecimal,
+    val status: Transaction.Status,
+    val senderId: UUID,
+    val senderName: String,
+    val receiverId: UUID,
+    val receiverName: String,
+    val type : String
+)
```
Karrar-Mohammed: what is the difference between this class and the Transaction class ?

wallet/src/main/kotlin/net/thechance/wallet/service/TransactionService.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.stereotype.Service
+import java.time.LocalDateTime
+import java.util.*
+
+@Service
+class TransactionService(
+    private val transactionRepository: TransactionRepository
+) {
+    fun getFilteredTransactions(
+        filter: TransactionFilterRequest,
```
Karrar-Mohammed: don't pass something from the api to the service, this is violate our architecture

wallet/src/main/resources/db-wallet/changelog/changelogs/changelogs_001.yaml
```diff
@@ -50,4 +50,56 @@ databaseChangeLog:
...
+        - onFail: MARK_RAN
+        - columnExists:
+            tableName: transactions
+            schemaName: wallet
+            columnName: type
+      changes:
+        - dropColumn:
+            tableName: transactions
+            schemaName: wallet
+            columnName: type
```
Karrar-Mohammed: why are we adding the column "type" then in the next query we remove it ?

Karrar-Mohammed:
as i see in the response we are returning many details of the page that we don't need like the sort, so i suggest we create our own page response class that contain only the details that we need to return

wallet/src/main/kotlin/net/thechance/wallet/repository/TransactionRepository.kt
Malak187: I don't think ORM can handle all of these cases, especially some cases like deciding whether to return the receiver name or dukan name based on the transaction type

wallet/src/main/kotlin/net/thechance/wallet/repository/transaction/TransactionProjection.kt
Malak187: Transaction class doesn't have the type variable as it is not reasonable to store type of transaction in our database. Also, Transaction class doesn't have info about sender or receiver rather than their IDs, so I needed to make a projection that let me expose only what I want, independent of the DB schema. I saw many examples following that approach.

wallet/src/main/resources/db-wallet/changelog/changelogs/changelogs_001.yaml
Malak187: Because At first I added the "type" column to the entity and then realized that it is doesn't make sense to store transaction "type" in the database so i removed it

wallet/src/main/kotlin/net/thechance/wallet/repository/TransactionRepository.kt
Al-Qassim: we think the query is too complicated to be handeled by the orm

wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/TransactionResponse.kt
Malak187: it's necessary so that the transaction is correctly defined, yes the ui wont need it, but the business logic may need it in the mobile for future features

wallet/src/main/resources/db-wallet/changelog/changelogs/changelogs_001.yaml
```diff
@@ -50,4 +50,36 @@ databaseChangeLog:
...
+              - column:
+                  name: status
+                  type: VARCHAR(50)
+                  defaultValue: 'SUCCESS'
+                  constraints:
+                    nullable: false
+
+
+  - changeSet:
+      id: 6-wallet-user
```
Al-Qassim: `6-wallet`

Al-Qassim (approved):

EL-MANCY (approved):

wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/TransactionMapper.kt
```diff
@@ -0,0 +1,26 @@
...
+fun Transaction.toResponse(currentUserId: UUID): TransactionResponse {
+    val userType = when (type) {
+        Transaction.Type.P2P -> {
+            if (sender.userId == currentUserId) UserTransactionType.SENT
+            else UserTransactionType.RECEIVED
+        }
+        Transaction.Type.ONLINE_PURCHASE -> UserTransactionType.ONLINE_PURCHASE
+    }
+
+    val actualReceiverName = receiver.dukanName?.ifBlank { receiver.userName } ?: receiver.userName
```
Al-Qassim:
```
val actualReceiverName = 
    if (userType == UserTransactionType.ONLINE_PURCHASE) 
        receiver.dukanName.ifBlankOrNull { receiver.userName }
    else 
        receiver.userName
```

Malak187: makes sense. thanks

wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
```diff
@@ -0,0 +1,69 @@
...
+            transactions = transactions.content,
+            page = transactions.number,
+            pageSize = transactions.size,
+            totalElements = transactions.totalElements,
+            totalPages = transactions.totalPages
+        )
+        return ResponseEntity.ok(response)
+    }
+
+    @GetMapping("/first")
```
MuhammedEdrees: Naming could be improved to indicate that the endpoint is for the first date not the first transaction

MuhammedEdrees (approved):
Well done 👏

wallet/src/main/kotlin/net/thechance/wallet/entity/Transaction.kt
```diff
@@ -15,11 +16,21 @@ data class Transaction(
...
+    @Column(nullable = false)
+    val type: Type,
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "sender_id", nullable = false)
+    val sender: WalletUser,
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "receiver_id", nullable = false)
+    val receiver: WalletUser,
```
Karrar-Mohammed: i think using FetchType.LAZY here is not the best, because we usually will access the send and the receiver so in my opinion since we will access the sender and/or receiver details in many cases it's better to set it as FetchType.EAGER

Karrar-Mohammed (approved):
good job

PR #54, opened 2025-09-25
Feature/dkn 55 implement an end point that deletes a shelf

ElsayedMagdy122:
- Implement dukan shelf deletion endpoint

endpoint : 
/dukan/shelf/{shelfId}

### DIFF
--- a/dukan/src/main/kotlin/net/thechance/dukan/exception/ShelfDeletionNotAllowedException.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/exception/ShelfDeletionNotAllowedException.kt
@@ -1,0 +1,5 @@
+package net.thechance.dukan.exception
+
+class ShelfDeletionNotAllowedException(
+    override val message: String = "Shelf contains products and cannot be deleted"
+) : RuntimeException(message)
--- a/dukan/src/main/kotlin/net/thechance/dukan/exception/ShelfNotFoundException.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/exception/ShelfNotFoundException.kt
@@ -1,0 +1,3 @@
+package net.thechance.dukan.exception
+
+class ShelfNotFoundException (override val message: String = "Shelf not found"): Exception(message)
--- a/dukan/src/main/kotlin/net/thechance/dukan/repository/DukanShelfRepository.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/repository/DukanShelfRepository.kt
@@ -8,4 +8,5 @@ import java.util.UUID
 @Repository
 interface DukanShelfRepository : JpaRepository<DukanShelf, UUID> {
     fun existsByTitleAndDukanId(title: String, dukanId: UUID): Boolean
+    fun findByIdAndDukanId(id: UUID, dukanId: UUID): DukanShelf?
 }
--- a/dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
@@ -1,5 +1,6 @@
 package net.thechance.dukan.service
 
+import net.thechance.dukan.entity.Dukan
 import net.thechance.dukan.entity.DukanShelf
 import net.thechance.dukan.exception.DukanNotFoundException
 import net.thechance.dukan.exception.ShelfDeletionNotAllowedException
@@ -14,11 +15,11 @@ import java.util.UUID
 @Service
 class DukanShelfService(
     private val dukanShelfRepository: DukanShelfRepository,
-    private val dukanRepository: DukanRepository,
+    private val dukanService: DukanService,
     private val dukanProductRepository: DukanProductRepository
 ) {
     fun createShelf(title: String, ownerId: UUID): DukanShelf {
-        val dukan = dukanRepository.findByOwnerId(ownerId) ?: throw DukanNotFoundException()
+        val dukan = dukanService.getDukanByOwnerId(ownerId)
 
         if (dukanShelfRepository.existsByTitleAndDukanId(title, dukan.id)) {
             throw ShelfNameAlreadyTakenException()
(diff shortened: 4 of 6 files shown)


dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
```diff
@@ -27,4 +30,17 @@ class DukanShelfService(
             )
         )
     }
+    fun deleteShelf(shelfId: UUID, ownerId: UUID) {
+        val dukan = dukanRepository.findByOwnerId(ownerId) ?: throw DukanNotFoundException()
+
+        val shelf = dukanShelfRepository.findByIdAndDukanId(shelfId, dukan.id)
+            ?: throw DukanNotFoundException()
```
haideraljawasim: I think you could here use `findById` only no need to dukan id since the id in shelf table represent the primary key and must be unique  , another thing i think you meant to throw `ShelfNotFoundException()` and not `DukanNotFoundException()`

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanShelfRepository.kt
```diff
@@ -8,4 +8,5 @@ import java.util.UUID
 @Repository
 interface DukanShelfRepository : JpaRepository<DukanShelf, UUID> {
     fun existsByTitleAndDukanId(title: String, dukanId: UUID): Boolean
+    fun findByIdAndDukanId(id: UUID, dukanId: UUID): DukanShelf?
 }
```
haideraljawasim, line 12: here what I'm saying no need to add this function just use `findById`

haideraljawasim (commented):
Great job you did here  👏 ,  Elsayed!

dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
ElsayedMagdy122: user can delete a shelf from their Dukan only

image

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanShelfRepository.kt
ElsayedMagdy122: same

dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
ElsayedMagdy122: yes, i refactored it , to ShelfNotFoundException

haideraljawasim: hmmm, now it make more sense, good job , so just change the exception to throw `ShelfNotFoundException()`

haideraljawasim (approved):

dukan/src/main/kotlin/net/thechance/dukan/exception/ShelfDeletionNotAllowedException.kt
```diff
@@ -0,0 +1,5 @@
+package net.thechance.dukan.exception
+
+class ShelfDeletionNotAllowedException(
+    override val message: String = "Shelf contains products and cannot be deleted"
+) : RuntimeException(message)
```
devfalah, line 5: I noticed this class uses `RuntimeException` but `ShelfNotFoundException` extends `Exception`
could you clarify the reson behind the difference?

dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
```diff
@@ -27,4 +31,17 @@ class DukanShelfService(
             )
         )
     }
+    fun deleteShelf(shelfId: UUID, ownerId: UUID) {
+        val dukan = dukanRepository.findByOwnerId(ownerId) ?: throw DukanNotFoundException()
```
devfalah: This logic is repeated I suggest creating a function in DukanService that retrieves a dukan by ID and throws this exception

dukan/src/main/kotlin/net/thechance/dukan/exception/ShelfNotFoundException.kt
```diff
@@ -0,0 +1,3 @@
+package net.thechance.dukan.exception
+
+class ShelfNotFoundException (override val message: String = "Shelf not found"): Exception(message)
```
TarekIdrees, line 3: I would say having separate files for each entity's exceptions would be nice. What do you think? 
- DukanExceptios 
- ProductExceptions 
- ShelfExceptions

ElsayedMagdy122: I am glad to hear your comment, I agree with you, we will open a PR to make the Error handling fixed in the same way for all of us ❤

dukan/src/main/kotlin/net/thechance/dukan/exception/ShelfDeletionNotAllowedException.kt
ElsayedMagdy122: Thanks for pointing this out 👍. The reason is that Exception is checked and forces explicit handling, 
 while RuntimeException is unchecked. In Spring Boot we usually prefer extending RuntimeException so that 
 errors can bubble up and be handled in a centralized way with @ControllerAdvice) 
 instead of forcing try/catch blocks everywhere.

dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
ElsayedMagdy122: Great suggestion 👍. You're right

devfalah (approved):

mostafa-n3ma (approved):
nice 👍🏻
will discuss it further latter on ...

TarekIdrees (approved):
Nice work!

PR #55, opened 2025-09-25
Implement reset password

ThoriaHamdy99:
## what is the PR Scope?

The scope is adding reset password service with IP rate limit.

## why do we need that?

- `/reset-password` to allow user to reset his/her password after opt verification was done

- Endpoint: `/identity/authentication/reset-password`

Curl:
```
curl --location 'https://mena-dev.the-chance.net/identity/authentication/reset-password' \
--header 'Content-Type: application/json' \
--data '{
    "newPassword":"12345679",
    "confirmPassword":"12345679",
    "sessionId": "9e7c7df5-0dcc-4525-8ffa-601aeeee7e17"
}'
```

RequestBody: 
```json
{
    "newPassword":"12345679",
    "confirmPassword":"12345679",
    "sessionId": "9e7c7df5-0dcc-4525-8ffa-601aeeee7e17"
}
```

Expected returned body:
```json
{
    "message": "Password reset successfully"
}
```


identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -18,4 +18,19 @@ class UserService(
...
     }
+
+    fun updatePasswordByPhoneNumber(phoneNumber: String, newPassword: String): Boolean {
+        val userWithNewPassword = getUserWithNewPassword(phoneNumber, newPassword)
+        return try {
+            val savedUser = userRepository.save(userWithNewPassword)
+            savedUser.password == newPassword
+        } catch (exception: Exception) {
+            false
+        }
```
poula22: If an exception happen in this case, it will be hard to know what is the root cause of it and also this catch  block ignores the role of ControllerAdvice as exception handler

identity/src/test/kotlin/net/thechance/identity/service/ResetPasswordServiceTest.kt
```diff
@@ -0,0 +1,87 @@
...
+
+class ResetPasswordServiceTest {
+    private val userService: UserService = mockk(relaxed = true)
+    private val passwordEncoder: PasswordEncoder = mockk(relaxed = true)
+    private val resetPasswordService: ResetPasswordService by lazy {
+        ResetPasswordService(
+            userService = userService,
+            passwordEncoder = passwordEncoder
+        )
+    }
```
poula22: What is the value of lazy block here?

identity/src/test/kotlin/net/thechance/identity/service/ResetPasswordServiceTest.kt
```diff
@@ -0,0 +1,87 @@
...
+        assertThrows(PasswordNotUpdatedException::class.java) {
+            resetPasswordService.resetPassword(phoneNumber, newPassword, confirmPassword)
+        }
+    }
+}
+
+private const val phoneNumber = "+201145236258"
+private const val newPassword = "12345678"
+private const val confirmPassword = "12345678"
+private const val wrongConfirmPassword = "12345679"
```
poula22: I have a suggestion here we have util package with dummy phone numbers, passwords, users,etc. we can add to them these constants. So If we need in the future to make test cases for more complex logic, the util package will help us with that.

poula22 (commented):
Great work @ThoriaHamdy99👏. There is only 2 small enhancement comments.

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
ThoriaHamdy99: here I need to know if the user is updated succcessfully or not.. If any exception happend so the user not updated and this all what I need to know and what I need to return to the service and it will handle its exception

identity/src/test/kotlin/net/thechance/identity/service/ResetPasswordServiceTest.kt
ThoriaHamdy99: I already used the dummy in the util package. these dummy variables for my test casses only instead of using dumyUsers.valideUser I made a readable variables for my test cases to be used.

identity/src/test/kotlin/net/thechance/identity/service/ResetPasswordServiceTest.kt
ThoriaHamdy99: I think it doesn't add any value here.. I will remove it

ThoriaHamdy99: Removed @poula22

identity/src/test/kotlin/net/thechance/identity/service/ResetPasswordServiceTest.kt
ThoriaHamdy99: I understood you based on our discussion, and I handled them, please check changes @poula22

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
ThoriaHamdy99: As per our discussion I updated it so please check again @poula22

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
```diff
@@ -0,0 +1,19 @@
...
+class ResetPasswordService(
+    private val userService: UserService,
+    private val passwordEncoder: PasswordEncoder
+) {
+    fun resetPassword(phoneNumber: String, newPassword: String, confirmPassword: String) {
+        if (newPassword != confirmPassword) throw PasswordMismatchException()
+        val encodedPassword = passwordEncoder.encode(newPassword)
+        val isPasswordUpdated = userService.updatePasswordByPhoneNumber(phoneNumber, encodedPassword)
+        if (!isPasswordUpdated) throw PasswordNotUpdatedException()
+    }
```
Karrar-Mohammed: i think it's better we also send the current password so we can make sure the user itself is changing the password

ThoriaHamdy99: It is nice idea and if not changed I can send a failure that this password already exists or something like that, right?

ThoriaHamdy99: But the mobile app does not know about the old password, I made a reset here and the user enter the new password and confirm password only

Karrar-Mohammed: oh ok so this is for reset the password if the user forget his password ?

Karrar-Mohammed: it that the case then i think the current one is enough

Karrar-Mohammed:
why this PR is draft ? is it finished or you still working on it ?

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
ThoriaHamdy99: Yep, this is the case only.

ThoriaHamdy99:
> why this PR is draft ? is it finished or you still working on it ?

I am waiting this PR as I need to add a check for the session id and IP rate limit
https://github.com/TheChance101/MENA-backend/pull/63

AhmadKharfan (approved):

identity/src/main/kotlin/net/thechance/identity/repository/OtpLogRepository.kt
```diff
@@ -39,6 +41,18 @@ interface OtpLogRepository: JpaRepository<OtpLog, UUID> {
...
     )
 
+    @Modifying
+    @Transactional
+    @Query("""
+        UPDATE OtpLog o 
+        SET o.expireAt = :now
+        WHERE o.sessionId = :sessionId
+    """)
+    fun expireLatestOtpBySessionId(
```
eslam-99: I think it's better to be `expireOtpBySessionId` as only 1 entry has the `sessionId`

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
```diff
@@ -35,7 +37,29 @@ class ResetPasswordService(
         return VerifyOtpResponse("OTP verified successfully")
     }
 
+    fun resetPassword(newPassword: String, confirmPassword: String, sessionId: String) {
+        val latestOtp = getLatestNotExpiredOtp(sessionId)
+        if (latestOtp.sessionId.toString() != sessionId || !latestOtp.isVerified) throw UnauthorizedException()
+        if (newPassword != confirmPassword) throw PasswordMismatchException()
```
eslam-99, line 43: I suggest moving pass check first before database hits

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
```diff
@@ -35,7 +37,29 @@ class ResetPasswordService(
...
+        val encodedPassword = passwordEncoder.encode(newPassword)
+        val isPasswordUpdated = userService.updatePasswordByPhoneNumber(latestOtp.phoneNumber, encodedPassword)
+        if (!isPasswordUpdated) throw PasswordNotUpdatedException()
```
eslam-99: No need for extra check as `userService.updatePasswordByPhoneNumber()` already throws exception

identity/src/main/kotlin/net/thechance/identity/repository/OtpLogRepository.kt
ThoriaHamdy99: okay I will update it

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -3,26 +3,42 @@ package net.thechance.identity.service
...
+    }
 
-	fun userExists(userId: UUID): Boolean {
-		return userRepository.existsById(userId)
-	}
+    fun userExists(userId: UUID): Boolean {
+        return userRepository.existsById(userId)
+    }
+
+    fun updatePasswordByPhoneNumber(phoneNumber: String, newPassword: String): Boolean {
```
eslam-99: No need for returning boolean as the function already throws exception on failure

ThoriaHamdy99: But I check the returned user password if it is as expected or not and return Boolean. the catch just catches any unexpected exception happens from data base.

But based on your comment I will handle this function checks internally

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
ThoriaHamdy99: okay, it's another point of check that I did but I will handle it internally.

identity/src/main/kotlin/net/thechance/identity/service/OtpService.kt
```diff
@@ -32,10 +33,29 @@ class OtpService(
...
         otpLogRepository.verifyOtp(parsedSessionId)
     }
 
+    fun getLatestNotExpiredOtpBySessionId(sessionId: String): OtpLog {
+        val otpLog = getLatestOtpBySessionId(sessionId)
+        checkOtpExpiration(otpLog)
+        return otpLog
+    }
+
+    fun expireLatestOtpBySessionId(sessionId: String) {
```
Karrar-Mohammed: if the session is UUID why don't we pass it directly as UUID to the service ?

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
```diff
@@ -35,7 +37,29 @@ class ResetPasswordService(
         return VerifyOtpResponse("OTP verified successfully")
     }
 
+    fun resetPassword(newPassword: String, confirmPassword: String, sessionId: String) {
+        val latestOtp = getLatestNotExpiredOtp(sessionId)
+        if (latestOtp.sessionId.toString() != sessionId || !latestOtp.isVerified) throw UnauthorizedException()
```
Karrar-Mohammed: if we are getting the top by session id then how the session id of the otp can not equal to the session id that we passed to this function ?

ThoriaHamdy99: You are totally right. At first, I made I was getting it by phone number, so I made this check after that I removed the phone number from the request and forgot to remove this check, so I removed it now as Eslam reminds me on the above comment.

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
```diff
@@ -35,7 +37,29 @@ class ResetPasswordService(
...
+        }
+    }
+
     private fun checkPhoneNumberExistence(phoneNumber: String) {
-        userRepository.findByPhoneNumber(phoneNumber) ?: throw UserNotFoundException("User not found")
+        try {
+            userService.findByPhoneNumber(phoneNumber)
+        } catch (_: Exception) {
+            throw UserNotFoundException("User not found")
+        }
```
Karrar-Mohammed, line 64: i fount that the userService.findByPhoneNumber(phoneNumber) already throw exception if the user not found, then why do we need to catch the exception and then throw another exception ? how this can be benefit for us ?

identity/src/main/kotlin/net/thechance/identity/service/OtpService.kt
ThoriaHamdy99: Okay I will pass any session id to the otpService as UUID

Karrar-Mohammed (commented):
good job 👏 , just i need to know the answer for my questions before approving the pr

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
ThoriaHamdy99: The exception of the userService.findByPhoneNumber(phoneNumber) functions returns invalid credentials but I need to return user not found as customized exception to this case.

identity/src/main/kotlin/net/thechance/identity/api/controller/IdentityController.kt
```diff
@@ -50,4 +50,10 @@ class IdentityController(
         val response = resetPasswordService.verifyOtp(request.otp, request.sessionId)
         return ResponseEntity.ok(response)
     }
+
+    @PostMapping("/reset-password")
+    fun resetPassword(@RequestBody @Valid request: ResetPasswordRequest): ResponseEntity<String> {
+        resetPasswordService.resetPassword(request.newPassword, request.confirmPassword, request.sessionId)
+        return ResponseEntity.ok("Password reset successfully")
```
Karrar-Mohammed: no need to return string here just return ok 200 that will be enough, because return the string is not adding any value

identity/src/main/kotlin/net/thechance/identity/service/OtpService.kt
ThoriaHamdy99: I don't know if at the future we need to call function like verifyOtp() in the controller direct. in this case the controller will handle Parames before calling the service is this, okay?

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
Karrar-Mohammed: i think you can make the userService throw the UserNotFoundException because it more reasonable that the invalid credentials error, but it's ok you can do that later

identity/src/main/kotlin/net/thechance/identity/service/OtpService.kt
Karrar-Mohammed: as i understand you are getting the session id as string from the requester in the controller so the correct approach the controller will be the responsible to map the session id from string to UUID

identity/src/main/kotlin/net/thechance/identity/api/controller/IdentityController.kt
ThoriaHamdy99: Okay you are right as the client will add the user message based on the status. But it will not affect anything. I know that no need, but anyone used postman needs to check it will be readable message. 

will it affect the performance or something like that?

Karrar-Mohammed: no it will not effect the performance but this is how most of the people build api, there are some rules for each http method you can search about so it's like convention everyone used to follow, and for the postman he can check the status code if it 200 that mean the request is succeeded and no need for the message

identity/src/main/kotlin/net/thechance/identity/service/OtpService.kt
ThoriaHamdy99: Okay, got it so I will handle them from the controller side. Actually it will be easy to use as we don't need to map many times just it will be once at the controller

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
ThoriaHamdy99: Actually, you are right it is more reasonable I will align with whom wrote this and will be updated

identity/src/main/kotlin/net/thechance/identity/api/controller/IdentityController.kt
ThoriaHamdy99: Okay will be removed

identity/src/main/kotlin/net/thechance/identity/service/OtpService.kt
Karrar-Mohammed: exactly

ThoriaHamdy99: Done

identity/src/main/kotlin/net/thechance/identity/api/controller/IdentityController.kt
ThoriaHamdy99: Done

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
ThoriaHamdy99: About this it will affect another response so we can make it later not in this PR

ThoriaHamdy99:
> good job 👏 , just i need to know the answer for my questions before approving the pr

Thanks @Karrar-Mohammed for these useful comments🎉🎉, please check again I made changes

AhmadKharfan (approved):

eslam-99 (approved):

identity/src/main/kotlin/net/thechance/identity/service/ResetPasswordService.kt
Karrar-Mohammed: ok no problem

Karrar-Mohammed (approved):

PR #59, opened 2025-09-26
Feature: Get user's profile information

nabilBouzineDev:
## what is the PR Scope ?

- This PR implements fetching the current authenticated user's profile information.

## why do we need that ?

- We need it in order to allow the client access its own informations like (username, firstName, image ...)

`/profile/me`

```json
{
  "username": "nabilBz",
  "firstName": "Nabil",
  "lastName": "Bouzine",
  "profileImageUrl": "https://avatars.newone.com/u/64290059?v=4"
}
```


identity/src/main/kotlin/net/thechance/identity/api/controller/ProfileController.kt
```diff
@@ -0,0 +1,24 @@
...
+@RequestMapping("/profile")
+class ProfileController(
+	private val userService: UserService,
+) {
+
+	@GetMapping("/me")
+	fun getCurrentUserProfile(): ResponseEntity<ProfileResponse> {
+		val userId = SecurityContextHolder.getContext().authentication.let { UUID.fromString(it.name) }
+		val response = userService.findById(userId)
+		return ResponseEntity.ok(response)
```
devfalah: you can get the user ID this way
```suggestion
	fun getCurrentUserProfile(@AuthenticationPrincipal userId: UUID): ResponseEntity<ProfileResponse> {
		val response = userService.findById(userId)
		return ResponseEntity.ok(response)
```

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -1,21 +1,30 @@
...
-    }
+	fun findByPhoneNumber(phoneNumber: String): User {
+		return userRepository.findByPhoneNumber(phoneNumber) ?: throw InvalidCredentialsException("User not found")
+	}
 
-    fun userExists(userId: UUID): Boolean {
-        return userRepository.existsById(userId)
-    }
+	fun findById(userId: UUID): ProfileResponse {
+		return userRepository.findByIdOrNull(userId)?.toResponse()
```
devfalah: mapping to response should be done in the controller

identity/src/main/kotlin/net/thechance/identity/api/controller/ProfileController.kt
nabilBouzineDev: Thanks @devfalah, I used this method it looks more cleaner.

devfalah (approved):

ThoriaHamdy99 (commented):
Great work👌

ThoriaHamdy99 (approved):

AimanYosofiUlt (approved):

PR #60, opened 2025-09-27
fix getting product by id

haideraljawasim:
## what is the PR Scope ?
Fix the issue with getting a product by id , The problem was caused by the method name, which prevented JPA from generating the implementation
## why do we need that ?
cause there error building in backend deployment


ElsayedMagdy122 (approved):

ElsayedMagdy122:
Great Work bro 👏

iBareq (approved):

PR #62, opened 2025-09-27
#DKN-49 Create a product according to the selected shelf

TarekIdrees:
This PR mainly focuses on creating a product for the selected shelf

Endpoint: **_/dukan/product/create*

RequestBody: 
```
{
  "name": "Test Product 4",
  "description": "This is just a test product where where where bla bla bla bla bla",
  "price": 20.5,
  "shelfId": "d9f8694a-429b-49aa-975d-acd4933e428f"
}
```

Expected returned body :
```
{
    "productId": "1c854172-eaa5-481b-8e41-fc8edec9e004"
}
```

In case the product name is already taken: 
```
{
    "status": 400,
    "message": "Product name already taken"
}
```

In case the price is incorrect: 
```
{
    "status": 400,
    "message": "Product price must be greater than 0"
}
```

In case the description is out of interval: 
```
{
    "status": 400,
    "message": "Product description must be between 100 and 3000 characters"
}
```


dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanProductController.kt
```diff
@@ -26,4 +29,20 @@ class DukanProductController(
         )
         return ResponseEntity.ok(imageUrls)
     }
+
+    @PostMapping("/create/{shelfId}")
+    fun createProduct(
+        @PathVariable shelfId: UUID,
+        @AuthenticationPrincipal userId: UUID,
+        @RequestBody  request: ProductCreationRequest,
+    ): ResponseEntity<UUID> {
```
haideraljawasim: I think passing the shelf id  with the request body will be better cause no need to make as path variable ? what do you think

dukan/src/main/kotlin/net/thechance/dukan/api/dto/ProductCreationRequest.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.dukan.api.dto
+
+data class ProductCreationRequest(
+    val name: String,
+    val description: String,
+    val price: Double,
+)
```
haideraljawasim: i think we can use the advantage of validation here like    ` @field:NotBlank` this will validated name not blank it it simple to use  and will be aligned more with the rest of our code cause we already used it

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -19,19 +26,59 @@ class DukanProductService(
...
+        if (dukanProductRepository.existsByDukanIdAndNameIgnoreCase(dukan.id, name)) {
+            throw ProductNameAlreadyTakenException()
+        }
+        if (price <= 0.0) {
+            throw InvalidProductPriceException()
+        }
+        val descriptionRegex = Regex("^.{100,3000}$")
+        if (!descriptionRegex.matches(description)) {
+            throw InvalidProductDescriptionException()
+        }
```
haideraljawasim: if we used the annotations we will get rid from all this code and the if statements here

dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
```diff
@@ -45,4 +42,10 @@ class DukanShelfService(
 
         dukanShelfRepository.delete(shelf)
     }
+
+    fun getShelfById(shelfId: UUID, ownerId: UUID): DukanShelf {
```
haideraljawasim, line 50: can we have a better naming here cause this not exactly get the shelf by id it get the shelf by id for the logged in user

dukan/src/main/kotlin/net/thechance/dukan/exception/dukan_product/InvalidProductDescriptionException.kt
```diff
@@ -0,0 +1,3 @@
+package net.thechance.dukan.exception.dukan_product
+
+class InvalidProductDescriptionException: Exception("Product description must be between 100 and 3000 characters")
```
ElsayedMagdy122: I think if we use validation with Request Dto it will save all this

dukan/src/main/kotlin/net/thechance/dukan/exception/dukan_product/InvalidProductPriceException.kt
```diff
@@ -0,0 +1,3 @@
+package net.thechance.dukan.exception.dukan_product
+
+class InvalidProductPriceException(): Exception("Product price must be greater than 0")
```
ElsayedMagdy122: same

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
ElsayedMagdy122: @haideraljawasim  i agree with you

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanProductRepository.kt
```diff
@@ -9,4 +9,5 @@ import java.util.*
 interface DukanProductRepository : JpaRepository<DukanProduct, UUID> {
     fun findAllByShelfId(shelfId: UUID): List<DukanProduct>
     fun findAllByDukanId(dukanId: UUID): List<DukanProduct>
+    fun existsByDukanIdAndNameIgnoreCase(dukanId: UUID, name: String): Boolean
```
mostafa-n3ma, line 13: the name here is to long , what do mean by ignore case ?

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
mostafa-n3ma: me two ..☑️

mostafa-n3ma: and also we can specify the required exception with the validation

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanProductRepository.kt
TarekIdrees: The IgnoreCase keyword makes the comparison on that String property case-insensitive.

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanProductController.kt
TarekIdrees: As I know, if the parameter is not considered as a part of the Product entity, it is better to consider it as a path parameter

haideraljawasim: but the shelf id is a part of product entity , i mean we will use it to link the shelf at end

dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
TarekIdrees: fun createShelf(title: String, ownerId: UUID): DukanShelf
 fun deleteShelf(shelfId: UUID, ownerId: UUID)
 fun getShelfById(shelfId: UUID, ownerId: UUID): DukanShelf
   
 According to these functions, do you think we still ned to add the loggedInUser word to these functions?

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
TarekIdrees: Noted, I thought this was recommended by Karrar, for that reason I used this approach

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanProductController.kt
TarekIdrees: Noted, I can see that CretingDukan used the needed IDs within the requested body, so yes, I would do that.

mostafa-n3ma (approved):

haideraljawasim (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/dto/DukanProductCreateResponse.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.dukan.api.dto
+
+import java.util.UUID
+
+data class DukanProductCreateResponse(
```
devfalah:
```suggestion
data class DukanProductCreationResponse(
```

dukan/src/main/kotlin/net/thechance/dukan/exception/dukan_product/DukanProductCreationFailedException.kt
```diff
@@ -0,0 +1,4 @@
+package net.thechance.dukan.exception.dukan_product
+
+class DukanProductCreationFailedException: Exception("Product creation failed") {
+}
```
devfalah:
```suggestion
class DukanProductCreationFailedException: Exception("Product creation failed")
```

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -19,19 +27,49 @@ class DukanProductService(
...
+            throw e
         }
         dukanProductRepository.save(product.copy(imageUrls = imageUrls))
         return imageUrls
     }
 
+    fun createProduct(params: DukanProductCreationParams): UUID {
+        try {
+            val dukan = dukanService.getDukanByOwnerId(params.ownerId)
+            val shelf = dukanShelfService.getShelfById(params.shelfId, params.ownerId)
```
devfalah: you can use reference proxy (lazy-loaded entity):
val shelf = dukanShelfRepository.getReferenceById(params.shelfId)
since we only need a placeholder reference to pass it to the prodcut

devfalah (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/dto/DukanProductCreateResponse.kt
TarekIdrees: Noted!

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
TarekIdrees: Impressive point, thanks for sharing!

PR #66, opened 2025-09-28
Feature/TRN-39 upload trend endpoint

mohamedd-hassan:
## what is the PR Scope ?
- Add new endpoint to upload trend from user
- Add new exceptions for that new endpoint
- Refactor some file locations based on feedback to match with other modules
- Add Video storage service for saving the video to the S3 Bucket
- Increase Multipart File Size limit to 100MB from 1MB as the trends user story requires 100MB maximum videos
- Add `FileSizeFilter ` in app/config to get the max allowed size for requests from `application.properties`
- Add Default max file sizes for all modules in `application.properties`

## why do we need that ?
We need that for the upload trend functionality as part of the user story for publishing a new trend
we also need the file size filter because trends videos can be up to 100MB which other features may want something small like 5MB so we have a way to adjust per feature

Endpoint: Post to "trends/reels"
Request: Multipart video file
Response: 
```
data class UploadReelResponse(
    val reelId: UUID
)
```


trends/src/main/kotlin/net/thechance/trends/service/ReelsService.kt
```diff
@@ -62,4 +64,24 @@ class ReelsService(
...
+
+    fun uploadReel(currentUserId: UUID, file: MultipartFile): UUID {
+        val videoUrl = videoStorageService.uploadVideo(
+            file = file,
+            fileName = file.originalFilename ?: "Untitled",
+            folderName = TRENDS_FOLDER_NAME
+        )
+        val reel = Reel(
+            ownerId = currentUserId,
+            thumbnailUrl = "",
```
iZyadNasser: we can make the thumbnailUrl property nullable with a null default, so we don't have to pass it dumb value like that, we will only pass it in its endpoint when we need it

app/src/main/kotlin/net/thechance/config/FileSizeFilter.kt
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        filterChain.doFilter(request, response)
+    }
+
+    private fun getMaxSizeForPath(path: String): Long {
+        val module = path.trim('/').split('/').firstOrNull() ?: ""
+
+        val maxSizeMB = when (module) {
+            "dukan" -> dukanMaxSize
```
iZyadNasser: can't we make a general consts file for these values? so we have all the features names in one place

iZyadNasser (approved):

trends/src/main/kotlin/net/thechance/trends/service/ReelsService.kt
```diff
@@ -11,12 +11,14 @@ import org.springframework.data.domain.Pageable
...
 import org.springframework.transaction.annotation.Transactional
+import org.springframework.web.multipart.MultipartFile
 import java.util.*
 
 @Service
 class ReelsService(
     private val reelsRepository: ReelsRepository,
-    private val categoryRepository: CategoryRepository
+    private val categoryRepository: CategoryRepository,
+    private val videoStorageService: VideoStorageService,
```
ahmeddsamii, line 21: instead of injecting the videoStorageService here in reel service, what about injecting it in the controller directly? because it's already service and do the logic of uploading, what do you think?

trends/src/main/kotlin/net/thechance/trends/service/VideoStorageService.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val props: TrendsStorageProperties,
+) {
+    fun uploadVideo(
+        file: MultipartFile,
+        fileName: String,
+        folderName: String,
+    ): String {
+        val mimeType = file.contentType ?: throw InvalidVideoException()
+        val extension = allowedMimeTypes[mimeType] ?: throw InvalidVideoException()
+        try {
```
ahmeddsamii: You can use runCatching here and onFailure throw the exception because you did not do anything with the exception in the catch block, when i see the code i feel like it's java not kotlin 😂😂

trends/src/main/kotlin/net/thechance/trends/service/VideoStorageService.kt
```diff
@@ -0,0 +1,61 @@
...
+data class TrendsStorageProperties(
+    val bucket: String,
+    val cdnEndpoint: String
+)
+
+@Service
+@EnableConfigurationProperties(TrendsStorageProperties::class)
+class VideoStorageService(
+    private val trendsS3Client: S3Client,
+    private val props: TrendsStorageProperties,
```
ahmeddsamii: i think if you typed the full name it will be better

ahmeddsamii (commented):
You did a good job mohamed, just i've left some comments please check them 👏

trends/src/main/kotlin/net/thechance/trends/service/ReelsService.kt
mohamedd-hassan: because there some logic that happens in the `ReelService ` still like creating the object and saving it in the database in case of success. the `VideoStorageService` is closer to a repository but for the S3 storage

ahmeddsamii (approved):

app/src/main/kotlin/net/thechance/config/FileSizeFilter.kt
```diff
@@ -0,0 +1,68 @@
...
+    private var trendsMaxSize: Int = 5
+
+    @Value("\${identity-maxSize:5}")
+    private var identityMaxSize: Int = 5
+
+    @Value("\${faith-maxSize:5}")
+    private var faithMaxSize: Int = 5
+
+    @Value("\${chat-maxSize:5}")
+    private var chatMaxSize: Int = 5
```
Karrar-Mohammed: i suggest to create configuration properties class for these parameters

trends/src/main/kotlin/net/thechance/trends/api/dto/reel/UploadReelResponse.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.trends.api.dto.reel
+
+import java.util.UUID
+
+data class UploadReelResponse(
+    val reelId: UUID
+)
```
Karrar-Mohammed, line 7: why do we need to return the reelId ?

trends/src/main/kotlin/net/thechance/trends/entity/Reel.kt
```diff
@@ -12,12 +12,12 @@ data class Reel(
     val id: UUID = UUID.randomUUID(),
     @Column(name = "owner_id", nullable = false)
     val ownerId: UUID,
-    @Column(name = "thumbnail_url", nullable = false)
-    val thumbnailUrl: String,
+    @Column(name = "thumbnail_url", nullable = true)
+    val thumbnailUrl: String? = null,
```
Karrar-Mohammed, line 16: is it ok to save a video without a thumbnail ? and if it null what the mobile will display ? and how we will set the thumbnail ?

trends/src/main/kotlin/net/thechance/trends/api/dto/reel/UploadReelResponse.kt
ahmeddsamii: Because we need to rerurn it to mobile to complete the next step with this id to update the description and the categories

trends/src/main/kotlin/net/thechance/trends/entity/Reel.kt
ahmeddsamii: We will upload the thumbnail with other endpoint call based on reelId

trends/src/main/kotlin/net/thechance/trends/api/dto/reel/UploadReelResponse.kt
mohamedd-hassan: the mobile will take the id for the next step of the process to add the description and categories to it

Karrar-Mohammed (commented):
great job, i just need to get the answer for my question before approving the PR but overall it's good

trends/src/main/kotlin/net/thechance/trends/entity/Reel.kt
mohamedd-hassan: creating a thumbnail in the backend will use costly amazon services so we decided it since we extract it to display it on mobile anyway we will upload it to the backend so there is another endpoint @iZyadNasser will create to accept a thumbnail and add it to the reel

if for any reason that fails the mobile will display a failed placeholder instead

Karrar-Mohammed (approved):
good job 👏

PR #67, opened 2025-09-28
Feature/wlt 19 export transactions statement endpoint

MuhammedEdrees:
## what is the PR Scope ?
- Added endpoint for generating pdf statement of user transactions.

## why do we need that ?
- For the export transactions feature in the mobile client

Request: `GET /wallet/transactions/statement?type=ONLINE_PURCHASE&status=SUCCESS`
 
 Response: 

 | First page | Middle page | Last page |
|---------|---------|---------|
| First page| Middle page | Last page|


wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
```diff
@@ -0,0 +1,222 @@
...
+        endDate: LocalDate?,
+        type: UserTransactionType?,
+        status: Transaction.Status?,
+        response: HttpServletResponse
+    ) {
+        val earliestTransactionDate =
+            transactionRepository.findFirstBySender_UserIdOrReceiver_UserIdOrderByCreatedAtAsc(
+                userId,
+                userId
+            )?.createdAt ?: LocalDateTime.now()
```
Karrar-Mohammed: here you will make the query even if startDate is not null but you will not use it, so it's better to call repository when the startDate is null

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
```diff
@@ -0,0 +1,222 @@
...
+    private val templateEngine: TemplateEngine
+) {
+
+    fun generateStatementPdf(
+        userId: UUID,
+        startDate: LocalDate?,
+        endDate: LocalDate?,
+        type: UserTransactionType?,
+        status: Transaction.Status?,
+        response: HttpServletResponse
```
Karrar-Mohammed: i don't think it's good idea to write to the response or passing the response to the service this is violation to our architecture

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
```diff
@@ -0,0 +1,222 @@
...
+            endDate = endDateTime,
+            pageable = Pageable.unpaged(),
+            currentUserId = userId
+        )
+
+        if (transactions.isEmpty) {
+            response.status = HttpServletResponse.SC_NO_CONTENT
+            response.writer.write("No transactions found for the specified filters")
+            return
+        }
```
Karrar-Mohammed: this should not be in the service

Karrar-Mohammed: you can throw exception here and handle it in the error handler

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
```diff
@@ -0,0 +1,222 @@
...
+                ?: earliestTransactionDate
+
+        val endDateTime = endDate?.atTime(23, 59, 59, 59) ?: LocalDateTime.now()
+
+        val transactions = transactionRepository.findFilteredTransactions(
+            status = status,
+            transactionType = type?.name,
+            startDate = startDateTime,
+            endDate = endDateTime,
+            pageable = Pageable.unpaged(),
```
Karrar-Mohammed: is this will return all the transactions without paging ?

Karrar-Mohammed: if that the case then i don't think this is safe because it might to overflow the server memory if the transactions is too many

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
```diff
@@ -0,0 +1,222 @@
...
+        response.contentType = "application/pdf"
+        response.setHeader(
+            "Content-Disposition",
+            "attachment; filename=\"statement_${from.formatHeaderDate()}_to_${to.formatHeaderDate()}.pdf\""
+        )
+    }
+
+    private fun LocalDateTime.formatHeaderDate(): String = this.format(DateTimeFormatter.ofPattern("dd MMM yyyy"))
+    private fun LocalDateTime.formatRowItemDate(): String = this.format(DateTimeFormatter.ofPattern("yyyy/MM/dd"))
+    private fun LocalDateTime.formatRowItemTime(): String = this.format(DateTimeFormatter.ofPattern("HH:mm a"))
```
Karrar-Mohammed: as i understand all this logic is for formatting or generating the pdf, i don't see this should be in the service, i suggest to create a class that responsible for that and you pass the transactions

Karrar-Mohammed: in the future if we changed our way to generate the pdf then we need to change the service and the logic and i don't think this is a good idea

Karrar-Mohammed (changes requested):
good job, but we have points that need to be improved

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
MuhammedEdrees: Is it a better approach to write it page by page to the stream?

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
```diff
@@ -0,0 +1,222 @@
...
+        transactions: Page<Transaction>,
+        userId: UUID
+    ): Double {
+        return openingBalance + transactions.sumOf { if (userId == it.sender.userId) it.amount.unaryMinus() else it.amount }
+            .toDouble()
+    }
+
+    private fun mapTransactionForTemplate(currentUserId: UUID, transaction: Transaction): Map<String, Any> {
+        return mapOf(
+            "id" to "TX-" + transaction.id.toString().substring(0, 6).uppercase(),
```
Al-Qassim: we can make it 8 chars and allow lower case to have more possibilities 
`transaction.id.toString().substring(0, 8)`

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
Al-Qassim: I can think of two different ways to fix this.
1. we pass only `response.outputStream` as `stream: ServletOutputStream` to the service, and we can write the pdf in it.
2. we can limit the `Service` to only return an `htmlContent: String` and give it to the controller, the controller can then depend on another class (Converter & Writer) and give it the `htmlContent` and the `stream: ServletOutputStream` to generate the pdf and write it in the response

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
Al-Qassim: @MuhammedEdrees 
yes I think this will be better
load the first page of transactions from the repo, write it to the `htmlContent`, then the second page and write it, and so on.

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
MuhammedEdrees: I think the second way is more of what we need since the service wouldn't know anything related to controller/response. I'll try to implement it.

Al-Qassim (approved):

Ahmed7osny1 (approved):
great work mohammed

wallet/src/main/kotlin/net/thechance/wallet/service/helper/StatementData.kt
```diff
@@ -0,0 +1,18 @@
...
+    val userId: UUID,
+    val username: String,
+    val startDateTime: LocalDateTime,
+    val endDateTime: LocalDateTime,
+    val openingBalance: Double,
+    val closingBalance: Double,
+    val totalPages: Int,
+    val types: List<UserTransactionType>?,
+    val transactionProvider: (pageNum: Int) -> List<Transaction>
+)
```
Karrar-Mohammed, line 14: i don't like this approach i see it's complicated and not easy to follow, i suggest to create a class that in the api and depending on the service to fetch the transaction and responsible to generate the pdf, the idea is like the ui and view model, the controller get the request, then call this class let's call it the transaction pdf generator that is depend on the transaction to fetch the transaction page by page and generate the pdf, i think this approach will be simpler

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
```diff
@@ -0,0 +1,93 @@
...
+            username = username,
+            startDateTime = startDateTime,
+            endDateTime = endDateTime,
+            openingBalance = openingBalance,
+            closingBalance = closingBalance,
+            totalPages = firstPage.totalPages,
+            types = types,
+            transactionProvider = transactionProvider
+        )
+    }
```
Karrar-Mohammed: this is function too big i think if you follow the approach that i suggest it will be simpler

Karrar-Mohammed (approved):

PR #73, opened 2025-10-01
Feature/setup domain events

Karrar-Mohammed:
## what is the PR Scope ?

- Domain Events

## why do we need that ?

- implement a way to let the feature notify other features by using the event bus


devfalah (approved):

MuhammedEdrees (approved):

events/src/main/kotlin/net/thechance/events/identity/UserCreatedEvent.kt
```diff
@@ -0,0 +1,9 @@
+package net.thechance.events.identity
+
+import net.thechance.events.MenaEvent
+
+data class UserCreatedEvent(
+    val phoneNumber: String,
+    val firstName: String,
+    val lastName: String,
```
mohamedd-hassan, line 8: i think for the trends feature we will need the profile picture url as well if its possible here

haideraljawasim (approved):

events/src/main/kotlin/net/thechance/events/identity/UserCreatedEvent.kt
Joseph-Sameh-0, line 8: I think for chat feature we need the user Id because we get the user from it's id in the bearer token
the id from the token is from the identity user table
it should matches the our user table

Am I right or is there an another sol?

Karrar-Mohammed: this is just an example when the identity team will implement the sign up end point then they can update this event

Karrar-Mohammed: i answered this question in the comment above, this is just example not the final version

mohamedd-hassan: okay thanks

mohamedd-hassan (approved):

Joseph-Sameh-0 (approved):

PR #74, opened 2025-10-01
Fix/WLT-48 backend issues

Malak187:
## what is the PR Scope ?
Fixing some backend issues in wallet feature.

## why do we need that ?
- Mark the transaction type as `RECEIVED` when the current user is a dukan owner, not `ONLINE PURCHASE`.
- Apply appropriate sorting methods for transaction history (Descending) and for statement transactions (Ascending).
- Prevent showing the failed received transactions for the current user, only show the failed sent transactions.
- Refactor function names and move some files to their correct packages.

### DIFF
--- a/wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/TransactionMapper.kt
+++ b/wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/TransactionMapper.kt
@@ -1,6 +1,7 @@
 package net.thechance.wallet.api.dto.transaction
 
 import net.thechance.wallet.entity.Transaction
+import net.thechance.wallet.service.helper.UserTransactionType
 import java.util.*
 
 fun Transaction.toResponse(currentUserId: UUID): TransactionResponse {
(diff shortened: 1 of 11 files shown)


wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/TransactionMapper.kt
```diff
@@ -9,7 +10,10 @@ fun Transaction.toResponse(currentUserId: UUID): TransactionResponse {
             if (sender.userId == currentUserId) UserTransactionType.SENT
             else UserTransactionType.RECEIVED
         }
-        Transaction.Type.ONLINE_PURCHASE -> UserTransactionType.ONLINE_PURCHASE
+        Transaction.Type.ONLINE_PURCHASE -> {
+            if (sender.userId == currentUserId) UserTransactionType.ONLINE_PURCHASE
+            else UserTransactionType.RECEIVED
+        }
```
Karrar-Mohammed, line 16: why do we need that ?

Karrar-Mohammed (approved):

wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/TransactionMapper.kt
Karrar-Mohammed: ok got it

MuhammedEdrees (approved):

EL-MANCY (approved):

PR #75, opened 2025-10-01
Update all users in the database

nabilBouzineDev:
## what is the PR Scope ?
Add database migration scripts to insert test users  in the identity schema and fix the deployment

## why do we need that ?
This will allow us to test the profile feature easily with the same easier development and provided users with the known credentials.


poula22 (approved):

identity/src/main/resources/db-identity/changelog/changelogs/changelogs_002.yaml
```diff
@@ -2,6 +2,21 @@ databaseChangeLog:
...
+      changes:
+        - delete:
+            schemaName: identity
+            tableName: refresh_tokens
+        - delete:
+            schemaName: identity
+            tableName: login_log
+        - delete:
+            schemaName: identity
+            tableName: users
```
Karrar-Mohammed, line 15: will this delete the table or the records ?

Karrar-Mohammed: also i don't prefer to add these to the database migration i prefer to delete it directly from the database

nabilBouzineDev: This will delete the records

nabilBouzineDev: I agree with you @Karrar-Mohammed. 

But, there's some users already in the database and I can't  delete them manually because I don't have access to the depolyed server

Karrar-Mohammed: and what is the issue if we keep the current users ? why we need to delete them ?

poula22: @Karrar-Mohammed  We need to add new columns and one of them is unique, so we need to add a unique value to each user in the database

nabilBouzineDev: We can't keep the current users since we're adding new columns with constraints that will make Liquibase clueless.

For example:

Karrar-Mohammed (approved):

poula22 (approved):

eslam-99 (approved):

PR #76, opened 2025-10-02
Feature/dkn 61 error handling

ElsayedMagdy122:
## what is the PR Scope ?
error handling for dukan

## why do we need that ?
We need this for error handling, to standardize the response structure across the system and ensure that all errors return in a consistent format

{
  "message": "Shelf not found",
  "errors": null,
  "errorCode": 1003
}


dukan/src/main/kotlin/net/thechance/dukan/exception/DukanShelfExceptions.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import org.springframework.http.HttpStatus
+
+@ErrorCode(code = 1401, status = HttpStatus.CONFLICT)
+class ShelfDeletionNotAllowedException(
+    override val message: String = "Shelf contains products and cannot be deleted"
+) : Exception(message)
+
+@ErrorCode(code = 1402, status = HttpStatus.NOT_FOUND)
+class ShelfNotFoundException(override val message: String = "Shelf not found") : Exception(message)
```
TarekIdrees: Since the ShelfDeletionNotAllowedException  and  ShelfNotFoundExceptio classes describe what is intended to say, there is no need to pass the message. 

Instead, we can place it directly in the Exception class as you did for Dukan and DukanProduct exceptions.

TarekIdrees (changes requested):

dukan/src/main/kotlin/net/thechance/dukan/exception/DukanShelfExceptions.kt
ElsayedMagdy122: @TarekIdrees i fixed it , thanks for notice❤

devfalah (approved):

TarekIdrees (approved):
You did well

mostafa-n3ma (approved):

PR #86, opened 2025-10-05
feature/WLT-61-add `get user details` endpoint 

Joseph-Sameh-0:
## what is the PR Scope ?
- The PR implements the /wallet/user/details endpoint, to get user details like name and image 
 
## why do we need that ?
- to know the name and the image that the user will pay for

### request
```HTTP
GET /wallet/user/details?userId={{uuid}}
Content-Type: application/json
Authorization: Bearer token
```

### response
```JSON
{
  "name": "Diana Miller",
  "ImageUrl": null
}
```
or if the user doesn't exist
```JSON
{
  "status": 500,
  "message": "User not found"
}
```


wallet/src/main/kotlin/net/thechance/wallet/api/controller/BalanceController.kt
```diff
@@ -24,4 +29,19 @@ class BalanceController(
 
         return ResponseEntity.ok(response)
     }
+
+    @PostMapping("/validate-payment-amount")
```
Malak187: I think for simplicity you can just call it validate-payment

wallet/src/main/kotlin/net/thechance/wallet/service/helper/PaymentAmountValidationResult.kt
```diff
@@ -0,0 +1,8 @@
+package net.thechance.wallet.service.helper
+
+data class PaymentAmountValidationResult(
+    val isValid: Boolean,
+    val currentBalance: Double,
+    val recipientName: String,
+    val recipientImageUrl: String? = null
+)
```
Malak187: just for consistency, replace every `recipient` word to `receiver` in the PR

wallet/src/main/kotlin/net/thechance/wallet/service/WalletService.kt
```diff
@@ -17,4 +20,18 @@ class WalletService(
...
+
+        val currentBalance = getUserBalance(userId)
+        val isValid = amount > 0 && amount <= currentBalance
+        return PaymentAmountValidationResult(
+            isValid = isValid,
+            currentBalance = currentBalance,
+            recipientName = recipient.firstName + " " + recipient.lastName,
+            recipientImageUrl = recipient.imageUrl
+        )
+    }
```
Malak187: do we need to ensure that the sender and receiver cannot be the same person? maybe if it is available for the user to send the amount to himself

Malak187 (commented):
Great work Joseph

wallet/src/main/kotlin/net/thechance/wallet/api/controller/BalanceController.kt
Joseph-Sameh-0: it's not the payment itself to name it validate-payment 
it just validate on the amount only
what do you think?

wallet/src/main/kotlin/net/thechance/wallet/service/WalletService.kt
Joseph-Sameh-0: I don't think that we have a requirement says that the user can't send to himself 😂

Malak187: but  it doesn't make sense, why wasting a request like that?

wallet/src/main/kotlin/net/thechance/wallet/api/controller/BalanceController.kt
Malak187: what about `/payments/confirm`? just like the design.
Also I suggest to make a payment controller since we'll add more functions for handling payments so it would be separated in another controller and service also

Malak187: I also read before that it is better that the naming of endpoint have to represent a resource like noun, not an action. You can check endpoints naming best practice

wallet/src/main/kotlin/net/thechance/wallet/service/WalletService.kt
Joseph-Sameh-0: I think we should ask before making this condition

wallet/src/main/kotlin/net/thechance/wallet/api/controller/BalanceController.kt
Joseph-Sameh-0: ok I will do it
note: it's not the payment confirmation

wallet/src/main/kotlin/net/thechance/wallet/entity/user/WalletUser.kt
```diff
@@ -16,5 +16,12 @@ data class WalletUser(
     @Column(nullable = false)
     val userName: String,
```
MuhammedEdrees, line 17: I think we might need to remove user name since we added first and last name to match identity's user. We don't have to do it on this pr but let's keep it in mind.

MuhammedEdrees (approved):
Well done 👏

Malak187 (approved):
Great work Joseph, just update the PR description after your changes

EL-MANCY (approved):
Great work Joe

wallet/src/main/kotlin/net/thechance/wallet/api/controller/UserController.kt
```diff
@@ -0,0 +1,25 @@
...
+
+
+@RestController
+@RequestMapping("/wallet/user")
+class UserController(
+    private val walletUserService: WalletUserService,
+) {
+    @GetMapping("/details")
+    fun getUserDetails(
+        @AuthenticationPrincipal currentUserId: UUID,
```
Malak187: Do we need currentUserId? I see that it is not used

Malak187 (approved):
Great work 👏

wallet/src/main/kotlin/net/thechance/wallet/api/controller/UserController.kt
Joseph-Sameh-0: UR right I will remove it

wallet/src/main/kotlin/net/thechance/wallet/api/controller/UserController.kt
MuhammedEdrees, line 1: When the payment is for online shopping should the returned data represent the owner of the dukan or the dukan itself? if it's the dukan does dukan has image that we can return or how will it be handled

Joseph-Sameh-0: what is your opinion about it ?
and is it in the current story?

MuhammedEdrees: I think in this case the dukan image and name should be returned to distinguish between the two transaction types.

Joseph-Sameh-0: @EL-MANCY

PR #92, opened 2025-10-08
implement endpoint to get dukans by categoryId

haideraljawasim:
## what is the PR Scope ?
This PR implement the logic and endpoint of getting Dukan's by category id as paged data 

## why do we need that ?
We need this cause one of the requirement is getting Dukan's under a certain category 

`/dukan/categorie/10a76d2d-ef76-4baf-bd7a-170a2e602f68?size=10&sort=createdAt,desc`

The id here represent a path variable which is the category id 

Response 
```
{
    "totalElements": 1,
    "totalPages": 1,
    "first": true,
    "last": true,
    "size": 10,
    "content": [
        {
            "id": "0efe45c9-a249-41a6-87b8-eb0f46c026f4",
            "name": "Dukan",
            "imageUrl": "https://unbenignly-unhurt-anne.ngrok-free.dev/dukan-bucket/images/dukan/Dukan-Dukandukan_image.png_2025-10-04T18:48:33.137547.png"
        }
    ],
    "number": 0,
    "sort": {
        "empty": false,
        "unsorted": false,
        "sorted": true
    },
    "pageable": {
        "pageNumber": 0,
        "pageSize": 10,
        "sort": {
            "empty": false,
            "unsorted": false,
            "sorted": true
        },
        "offset": 0,
        "unpaged": false,
        "paged": true
    },
    "numberOfElements": 1,
    "empty": false
}

```

### DIFF
--- a/dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
@@ -8,7 +8,12 @@ import net.thechance.dukan.mapper.DukanLanguage
 import net.thechance.dukan.mapper.toDto
 import net.thechance.dukan.mapper.toDukanCreationParams
 import net.thechance.dukan.mapper.toDukanStyleResponse
+import net.thechance.dukan.mapper.toResponse
 import net.thechance.dukan.service.DukanService
+import org.springframework.data.domain.Page
+import org.springframework.data.domain.Pageable
+import org.springframework.data.domain.Sort
+import org.springframework.data.web.PageableDefault
 import org.springframework.http.ResponseEntity
 import org.springframework.security.core.annotation.AuthenticationPrincipal
 import org.springframework.web.bind.annotation.*
@@ -24,7 +29,7 @@ class DukanController(
     @GetMapping("/styles")
     fun getAllStyles(): ResponseEntity<DukanStyleResponse> {
         val styles = dukanService.getAllStyles().toDukanStyleResponse()
-       return ResponseEntity.ok(styles)
+        return ResponseEntity.ok(styles)
     }
 
     @GetMapping("/categories")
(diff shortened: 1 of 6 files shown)


dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
```diff
@@ -67,12 +72,20 @@ class DukanController(
...
+
     @GetMapping("/statues")
     fun getDukanStatues(
         @AuthenticationPrincipal userId: UUID,
     ): ResponseEntity<DukanStatuesResponse> {
         val dukan = dukanService.getDukanByOwnerId(userId)
         return ResponseEntity.ok(DukanStatuesResponse(dukan.name, dukan.status))
     }
+
+    @GetMapping("/categorie/{categoryId}")
```
devfalah:
```suggestion
    @GetMapping("/{categoryId}")
```

devfalah (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanController.kt
haideraljawasim: I didn't write it like this cause this path already used to get Dukan details

ElsayedMagdy122 (approved):

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanRepository.kt
```diff
@@ -10,4 +12,6 @@ interface DukanRepository : JpaRepository<Dukan, UUID> {
     fun existsByName(name: String): Boolean
     fun existsByOwnerId(ownerId: UUID): Boolean
     fun findByOwnerId(ownerId: UUID): Dukan?
+
+    fun findAllByCategoriesId(categoryId: UUID, pageable: Pageable): Page<Dukan>
```
TarekIdrees, line 16: It would be better if it is categoryId instead of categoriesId

TarekIdrees (approved):
Good job 
Just fix this tiny typo 👏🏻

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanRepository.kt
haideraljawasim: we have the column name categories so if i named it  category  JPA wouldn't be able to map it correctly when implement it , that why i named it like this

TarekIdrees: Hmm, understood 
I don't know if there is any way to refer for the category_id which is inverse id of the category dukan

PR #97, opened 2025-10-10
Feature/cht 65 get user endpoint

omer1998:
## what is the PR Scope ?
add endpoint to get the data of the user 

## why do we need that ?
in order to display the data in required section

endpoint: chat/user
```
{
    "firstName": "osama",
    "lastName": "kamel",
    "phoneNumber": "+967775074564",
    "imageUrl": null
}
```

### DIFF
--- a/chat/src/main/kotlin/net/thechance/chat/api/dto/UserDto.kt
+++ b/chat/src/main/kotlin/net/thechance/chat/api/dto/UserDto.kt
@@ -1,16 +1,16 @@
-package net.thechance.chat.service.model
+package net.thechance.chat.api.dto
 
 import net.thechance.chat.entity.ContactUser
 
-data class UserModel(
+data class UserDto(
     val firstName: String,
     val lastName: String,
     val phoneNumber: String,
     val imageUrl: String? = null,
 )
 
-fun ContactUser.toModel(): UserModel{
-    return UserModel(
+fun ContactUser.toDto(): UserDto{
+    return UserDto(
         firstName = this.firstName,
         lastName = this.lastName,
         phoneNumber = this.phoneNumber,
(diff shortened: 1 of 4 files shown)


chat/src/main/kotlin/net/thechance/chat/service/model/UserModel.kt
```diff
@@ -0,0 +1,19 @@
+package net.thechance.chat.service.model
+
+import net.thechance.chat.entity.ContactUser
+
+data class UserModel(
```
devfalah: Rename this to UserDTO and move it to dto package

chat/src/main/kotlin/net/thechance/chat/api/dto/UserDto.kt
```diff
@@ -0,0 +1,19 @@
...
+)
+
+fun ContactUser.toDto(): UserDto{
+    return UserDto(
+        firstName = this.firstName,
+        lastName = this.lastName,
+        phoneNumber = this.phoneNumber,
+        imageUrl = this.imageUrl
+    )
+}
```
Al-Qassim, line 19: please move this mapper to a separate file called mapper

Al-Qassim (approved):

chat/src/main/kotlin/net/thechance/chat/api/dto/UserDto.kt
omer1998: right now, we are following our approach, we will do the refactor (as suggested by you) and definitely this issue will be addressed carefully

devfalah (approved):

raouf-kamel (approved):

PR #98, opened 2025-10-10
feature/DKN-75-editors-pick-endpoint

mostafa-n3ma:
## what is the PR Scope ?
dukan end points implementation 

## why do we need that ?

wee need an end point to return editor picks dukans pages but since we don;t have user preferences in the db right now we are returning all the valide dukans with at least on shelf and on product in it 
request : http://localhost:8080/dukan/editor_picks
response:
'''

{
  "totalElements": 2,
  "totalPages": 1,
  "first": true,
  "last": true,
  "size": 5,
  "content": [
    {
      "id": "22222222-2222-2222-2222-222222222222",
      "name": "My Test Dukan",
      "imageUrl": "https://picsum.photos/200/200?random=99"
    },
    {
      "id": "ca64095d-4e53-426d-bdc7-cff205b743a4",
      "name": "Demo Dukan",
      "imageUrl": "https://picsum.photos/200/200"
    }
  ],
  "number": 0,
  "sort": {
    "empty": false,
    "unsorted": false,
    "sorted": true
  },
  "pageable": {
    "pageNumber": 0,
    "pageSize": 5,
    "sort": {
      "empty": false,
      "unsorted": false,
      "sorted": true
    },
    "offset": 0,
    "unpaged": false,
    "paged": true
  },
  "numberOfElements": 2,
  "empty": false
}
'''

### DIFF
--- a/dukan/src/main/kotlin/net/thechance/dukan/repository/DukanRepository.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/repository/DukanRepository.kt
@@ -4,6 +4,7 @@ import net.thechance.dukan.entity.Dukan
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.Pageable
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
 import java.util.UUID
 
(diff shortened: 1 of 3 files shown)


haideraljawasim (approved):

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanRepository.kt
```diff
@@ -4,14 +4,33 @@ import net.thechance.dukan.entity.Dukan
...
+        SELECT s FROM DukanShelf s
+        WHERE s.dukan = d
+    )
+    AND EXISTS (
+        SELECT p FROM DukanProduct p
+        WHERE p.dukan = d
+    )
+    ORDER BY d.createdAt DESC
+    """
+    )
```
devfalah, line 34: We dont need to check whether the Dukan has shelves since having products means it already has shelves
```suggestion
SELECT DISTINCT d
FROM Dukan d
WHERE d.status = net.thechance.dukan.entity.Dukan.Status.APPROVED
  AND EXISTS (
    SELECT 1
    FROM DukanProduct p
    WHERE p.dukan = d
  )
ORDER BY d.createdAt DESC

```

mostafa-n3ma: I thought about this two and discussed it with haider and yousof but how about if the proudct hase been added wrongly with out shelfs

devfalah: then just update you checks to 
```suggestion
    @Query(
        """
    SELECT DISTINCT d
    FROM Dukan d
    WHERE d.status = net.thechance.dukan.entity.Dukan.Status.APPROVED
    AND EXISTS (
        SELECT 1 
        FROM DukanShelf s
        WHERE s.dukan = d
    )
    AND EXISTS (
        SELECT 1
        FROM DukanProduct p
        WHERE p.dukan = d
    )
    ORDER BY d.createdAt DESC
    """
    )
```

devfalah (approved):

ElsayedMagdy122 (approved):

PR #102, opened 2025-10-11
Feature/wlt 64 implement pay transaction end point

Joseph-Sameh-0:
## what is the PR Scope ?
- implements endpoint to get receiver details like name and image 
- add end point to add transaction until confirm it 
- add end point to submit the transaction

```HTTP
GET /wallet/transactions/{{transactionId}}/receiver-details
Content-Type: application/json
Authorization: Bearer token
```
### response
```JSON
{
  "name": "Diana Miller",
  "imageUrl": null
}
```
or if the transaction doesn't exist
```JSON
{
  "status": 500,
  "message": "Transaction with ID {{transactionId}} not found or access denied."
}
```
---
```HTTP
POST /wallet/transactions/initiate
Content-Type: application/json
Authorization: Bearer token
{
  "receiverId": "451e4d6c-0380-41ed-95e6-275793c404c6",
  "amount": 500
}
```

Response
```JSON
"f141d6d2-1fa1-4c37-a8bd-91a4228064f9"
```
---
```HTTP
POST /wallet/payment/{{transactionId}}/submit
Content-Type: application/json
Authorization: Bearer token
```
without response body


wallet/src/main/kotlin/net/thechance/wallet/entity/Transaction.kt
```diff
@@ -10,7 +10,7 @@ import java.util.*
 @Table(name = "transactions", schema = "wallet")
 data class Transaction(
     @Id
-    @GeneratedValue(strategy = GenerationType.UUID)
+    @Column(columnDefinition = "uuid", updatable = false, nullable = false)
```
Malak187, line 13: why did you remove the generation line?

wallet/src/main/kotlin/net/thechance/wallet/service/TransactionService.kt
```diff
@@ -48,4 +56,9 @@ class TransactionService(
         ) ?: throw EntityNotFoundException("Transaction with ID $transactionId not found or access denied.")
     }
 
+    fun createTransaction(transaction: CreatePendingTransactionParams): PendingTransaction {
```
Malak187: why the transaction is not of type PendingTransaction directly? why do you need CreatePendingTransactionParams

wallet/src/main/kotlin/net/thechance/wallet/service/helper/CreatePendingTransactionParams.kt
```diff
@@ -0,0 +1,25 @@
+package net.thechance.wallet.service.helper
+
+import net.thechance.wallet.entity.PendingTransaction
+import net.thechance.wallet.entity.Transaction
+import net.thechance.wallet.entity.user.WalletUser
+import java.util.*
+
+data class CreatePendingTransactionParams(
```
Malak187: as a date class, i think it would be better to rename it to PandingTransactionParams

wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/CreatePendingTransactionRequest.kt
```diff
@@ -0,0 +1,22 @@
+package net.thechance.wallet.api.dto.transaction
+
+import net.thechance.wallet.entity.Transaction
+import net.thechance.wallet.service.helper.CreatePendingTransactionParams
+import java.util.*
+
+data class CreatePendingTransactionRequest(
```
Malak187: same here, PendingTransactionRequest. and same question here, why didn't you make PendingTransactionRequest maps to PendingTransaction directly without the need of Params?

wallet/src/main/kotlin/net/thechance/wallet/api/controller/PaymentController.kt
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.web.bind.annotation.RequestParam
+import java.util.*
+
+@RestController
+@RequestMapping("/wallet/payment")
+class PaymentController(
+    private val paymentService: PaymentService
+) {
+    @PostMapping("/submit")
+    fun validatePayment(
```
Malak187: If this function is responsible for direct payments, it would be clearer to rename it to pay or send/submitTransaction instead of validate. Based on the endpoint name (submit), it seems to represent the actual payment process, even if there’s some validation involved along the way

wallet/src/main/kotlin/net/thechance/wallet/service/PaymentService.kt
```diff
@@ -0,0 +1,82 @@
...
+            .orElseThrow { IllegalArgumentException("transaction not found") }
+
+        if (pendingTransaction.sender.userId != userId)
+            throw IllegalArgumentException("User is not authorized to pay this transaction")
+
+        if (userId == pendingTransaction.receiver.userId)
+            throw IllegalArgumentException("Sender and receiver cannot be the same")
+
+        if (walletService.getUserBalance(userId) < pendingTransaction.amount.toDouble())
+            throw IllegalArgumentException("Insufficient balance")
```
Malak187: What do you think about moving all payment validation logic into a private function, for example:
`private fun validatePayment(transactionId: UUID, userId: UUID)`
This would keep the main function more focused on the payment flow itself while moving the conditional checks to a separate method.

wallet/src/main/kotlin/net/thechance/wallet/service/PaymentService.kt
```diff
@@ -0,0 +1,82 @@
...
+
+        val block = getBlock()
+        val transaction = pendingTransaction.toTransaction(block)
+        transactionRepository.save(transaction)
+        pendingTransactionRepository.deleteById(transactionId)
+    }
+
+    private fun getBlock(): Block {
```
Malak187: maybe getCurrentBlock would be clearer

Malak187 (commented):
Great work Joseph.

wallet/src/main/kotlin/net/thechance/wallet/entity/Transaction.kt
Joseph-Sameh-0: because I need to set it 
when it's generated I can't set it's value

wallet/src/main/kotlin/net/thechance/wallet/service/TransactionService.kt
Joseph-Sameh-0: because the entity has different types of the parameters like the sender and receiver of type WalletUser and the amount is BigDecimal I will not passing it like this from the controller

wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/CreatePendingTransactionRequest.kt
Joseph-Sameh-0: the request doesn't have the userId and the service shouldn't depend on the request

wallet/src/main/kotlin/net/thechance/wallet/api/controller/PaymentController.kt
Joseph-Sameh-0: you are right I forgot to rename it

wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/CreatePendingTransactionRequest.kt
Joseph-Sameh-0: I will refactor it to use params directly

Malak187 (approved):
💯

EL-MANCY (approved):
Great Job ya Joe

wallet/src/main/kotlin/net/thechance/wallet/entity/WalletDukan.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import jakarta.persistence.*
+import net.thechance.wallet.entity.user.WalletUser
+import java.util.*
+
+@Entity
+@Table(name = "dukan", schema = "wallet")
+data class WalletDukan(
+    @Id
+    val userId: UUID,
```
MuhammedEdrees: We can replace with dukanId to integrate it with dukan events in case dukan data is changed.

MuhammedEdrees (approved):
3aaaaash 👏

Malak187 (approved):
Well done Joseph 👏👏

wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
```diff
@@ -124,7 +123,23 @@ class TransactionController(
...
     }
 
+    @PostMapping("/add")
+    fun createTransaction(
+        @AuthenticationPrincipal userId: UUID,
+        @RequestBody params: PendingTransactionParams,
+    ): ResponseEntity<UUID> {
+        val transaction = transactionService.createTransaction(params, userId)
+        return ResponseEntity.ok(transaction.id)
+    }
```
Karrar-Mohammed: who will call this end point?

wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
```diff
@@ -124,7 +123,23 @@ class TransactionController(
...
+
+    @GetMapping("/receiver-details")
```
Karrar-Mohammed: i'm thinking of making the path of this end point like this
 "wallet/transactions/{transactionId}/receiver-details"
 what do you think ?

wallet/src/main/kotlin/net/thechance/wallet/entity/PendingTransaction.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    @Column(nullable = false, updatable = false)
+    val amount: BigDecimal,
+
+    )
+
+
+fun PendingTransaction.toTransaction(
+    block: Block,
+    status: Transaction.Status = Transaction.Status.SUCCESS
```
Karrar-Mohammed: i prefer that we remove the default value in case someone doesn't know about this parameter and it will be mapped to SUCCESS which not what we need

Karrar-Mohammed: so the idea is not minimize the human errors

wallet/src/main/kotlin/net/thechance/wallet/entity/user/WalletUser.kt
```diff
@@ -16,5 +14,13 @@ data class WalletUser(
...
-)
+    @Column(nullable = false)
+    val firstName: String,
+    @Column(nullable = false)
+    val lastName: String,
+    @Column(nullable = true)
+    val imageUrl: String?,
+
+    @OneToOne(mappedBy = "user", cascade = [CascadeType.ALL], fetch = FetchType.EAGER)
+    val dukan: WalletDukan? = null
```
Karrar-Mohammed, line 26: here we have dukan in the user and we have user in the dukan, do we need that ? i think we might only need one of them

wallet/src/main/kotlin/net/thechance/wallet/service/PaymentService.kt
```diff
@@ -0,0 +1,48 @@
...
+        val pendingTransaction = validateAndGetTransaction(userId, transactionId)
+
+        val block = blockService.getCurrentBlock()
+        val transaction = pendingTransaction.toTransaction(block)
+        transactionRepository.save(transaction)
+        pendingTransactionRepository.deleteById(transactionId)
+    }
+
+    private fun validateAndGetTransaction(userId: UUID, transactionId: UUID): PendingTransaction {
+        if (transactionRepository.findByIdOrNull(transactionId) != null)
```
Karrar-Mohammed: if you only need to know if the transaction is exist or not then use the function existsById instead of the findById

wallet/src/main/kotlin/net/thechance/wallet/service/TransactionService.kt
```diff
@@ -48,4 +57,22 @@ class TransactionService(
         ) ?: throw EntityNotFoundException("Transaction with ID $transactionId not found or access denied.")
     }
 
+    fun createTransaction(transaction: PendingTransactionParams, senderId: UUID): PendingTransaction {
```
Karrar-Mohammed: i think you can add the senderId to the PendingTransactionParams

Karrar-Mohammed: also rename the parameter "transaction" to "transactionParams" or "initiateTransactionParams" to avoid the ambiguity in the names

Karrar-Mohammed: also change the createTransaction name in every place and make it initiateTransaction, because here are just initiating the transaction, create transaction might someone understand it as the functionality to create and finalize the transaction

wallet/src/main/kotlin/net/thechance/wallet/service/TransactionService.kt
```diff
@@ -48,4 +57,22 @@ class TransactionService(
...
+            throw IllegalArgumentException("Sender and receiver cannot be the same.")
+
+        val sender = walletUserRepository.getReferenceById(senderId)
+        val receiver = walletUserRepository.getReferenceById(transaction.receiverId)
+        return pendingTransactionRepository.save(transaction.toPendingTransaction(sender, receiver))
+    }
+
+    fun getTransactionReceiverDetails(transactionId: UUID): ReceiverDetails {
+        val pendingTransaction = pendingTransactionRepository.findByIdOrNull(transactionId)
+        val transaction = transactionRepository.findByIdOrNull(transactionId)
```
Karrar-Mohammed: here you are always makes two queries, but not always you need both so you should only call the other one when you need it

Karrar-Mohammed: also do we need to get the receiver details in some cases for the pending transaction and in other cases for the completed transaction ?

because i'm thinking if the transaction is completed then we may have end point to get the transaction details that will include the receiver details, 

i'm thinking maybe we only make this function to get the pending transaction receiver details

wallet/src/main/kotlin/net/thechance/wallet/service/helper/ReceiverDetails.kt
```diff
@@ -0,0 +1,19 @@
+package net.thechance.wallet.service.helper
```
Karrar-Mohammed: why do we have the helper package, all the classes inside this package is just a model classes so i think we can put it in the entity package

wallet/src/test/kotlin/net/thechance/wallet/service/BlockServiceTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        every { transactionRepository.countAllByBlockId(blockId) } returns 15L
+        every { transactionRepository.getAllByBlockId(eq(blockId), any()) } returns listOf()
+        every { blockRepository.save(any()) } answers { firstArg() }
+
+        val result = blockService.getCurrentBlock()
+
+        assertNotNull(result)
+        assertNotEquals(latestBlock, result)
+        assertEquals(64, result.previousBlockHash.length)
+        verify { blockRepository.save(any()) }
```
Karrar-Mohammed, line 106: don't make many assertion in one unit test, instead separate it into multiple unit tests

Karrar-Mohammed (commented):
good job, but there are some points that we can improve

wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
Joseph-Sameh-0: dukan feature, It's just temporarily until we get how the dukan stores the items and the transaction to update ours with the events and remove this end point

wallet/src/main/kotlin/net/thechance/wallet/entity/user/WalletUser.kt
Joseph-Sameh-0: I used the dukan in the user only
yes I can remove the other because it's not used

wallet/src/main/kotlin/net/thechance/wallet/service/TransactionService.kt
Joseph-Sameh-0: I get the PendingTransactionParams from the user
but the senderId from the auth 
If I added it to the PendingTransactionParams how I will know the sender id to add it in the request 
I have the token only

I can Add it but making another class PendingTransactionRequest and map it to PendingTransactionParams
what do you think ?

wallet/src/main/kotlin/net/thechance/wallet/service/TransactionService.kt
Joseph-Sameh-0: we need it for the pending only 
but the function called `getTransactionReceiverDetails` so if any one use it he will expected to get the transaction detail no matters it's pending or not right ?

wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
Joseph-Sameh-0: I don't understand what is the point of doing this ?
we always make the parameters at the end 
so why to make it like this ?

wallet/src/main/kotlin/net/thechance/wallet/service/TransactionService.kt
Joseph-Sameh-0: I made it a little better now
image

events/src/main/kotlin/net/thechance/events/wallet/TransactionInitiatedEvent.kt
```diff
@@ -0,0 +1,16 @@
+package net.thechance.events.wallet
+
+import net.thechance.events.MenaEvent
+import java.util.UUID
+
+data class TransactionInitiatedEvent(
```
Karrar-Mohammed: i prefer to call it InitiateTransactionEvent because the listener will initiate the transaction not the publisher

wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/InitiateTransactionRequest.kt
```diff
@@ -0,0 +1,21 @@
...
+data class InitiateTransactionRequest(
+    val receiverId: UUID,
+    val amount: Double,
+)
+
+fun InitiateTransactionRequest.toInitiateTransactionParam(
+    senderId: UUID,
+): InitiateTransactionParams {
+    return InitiateTransactionParams(
+        type = Transaction.Type.P2P,
```
Karrar-Mohammed: i prefer to pass the type from the controller

wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
```diff
@@ -124,7 +123,23 @@ class TransactionController(
         response.setHeader("X-Statement-End-Date", metadata.endDate.toString())
     }
 
+    @PostMapping("/initiate")
```
Karrar-Mohammed, line 126: i suggest to specify in the name that this end point initiate P2P transaction so it can be easily to understand this end point for P2P only

like we can name it for example : "p2p/initiate"

Joseph-Sameh-0: I asked @EL-MANCY and he said make it just initiate because it is always p2p

Karrar-Mohammed (approved):
Good job 👏

wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
Karrar-Mohammed: i'm saying because it always p2p we need to specify that so when someone call the end point can understand this is only for the p2p

PR #104, opened 2025-10-11
Feature/cht 69 get chat by id endpoint

omer1998:
## what is the PR Scope ?
implement getting specific chat detail based on specific chat id

## why do we need that ?
in order to get some information about specific chat like chat name, imageUrl.

"chat/{chatId}"

```
{
    "id": "825265f7-7e30-4ac3-b9fb-16ba3869610e",
    "name": "Raouf Kamel",
    "requesterId": "451e4d6c-0380-41ed-95e6-275793c404c6",
    "imageUrl": null
}
```

in case of not found chat 
image

### DIFF
--- a/chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
+++ b/chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
@@ -76,6 +76,21 @@ class ChatController(
         )
     }
 
+    @GetMapping("/{chatId}")
+    fun getChatDetail(
+        @PathVariable chatId : String,
+        @AuthenticationPrincipal userId: UUID
+    ): ResponseEntity<ChatResponse>{
+        val id = UUID.fromString(chatId)
+        val chat = chatService.getChatById(id) ?: throw IllegalStateException("chat not found")
+        val otherUser = chat.users.firstOrNull { it.id != userId }
+        val contact =otherUser?.let{
+            contactService.getContactByOwnerIdAndContactUserId(userId, it.id)
+        }
+        return ResponseEntity.ok(chat.toResponse(userId,contact))
+
+    }
+
     companion object {
         const val QUEUE_MESSAGES = "/queue/messages"
     }
--- a/chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
+++ b/chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
@@ -3,7 +3,6 @@ package net.thechance.chat.api.controller
 import net.thechance.chat.api.dto.*
 import net.thechance.chat.service.ChatService
 import net.thechance.chat.service.ContactService
-import net.thechance.chat.service.model.toResponse
 import org.springframework.data.domain.Pageable
 import org.springframework.http.ResponseEntity
 import org.springframework.messaging.handler.annotation.MessageMapping
--- a/chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
+++ b/chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
@@ -3,6 +3,7 @@ package net.thechance.chat.api.controller
 import net.thechance.chat.api.dto.*
 import net.thechance.chat.service.ChatService
 import net.thechance.chat.service.ContactService
+import net.thechance.chat.service.model.toResponse
 import org.springframework.data.domain.Pageable
 import org.springframework.http.ResponseEntity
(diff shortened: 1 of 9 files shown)


chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
```diff
@@ -76,6 +76,21 @@ class ChatController(
         )
     }
 
+    @GetMapping("/{chatId}")
+    fun getChatDetail(
+        @PathVariable chatId : String,
+        @AuthenticationPrincipal userId: UUID
+    ): ResponseEntity<ChatResponse>{
+        val id = UUID.fromString(chatId)
+        val chat = chatService.getChatById(id) ?: throw IllegalStateException("chat not found")
```
Al-Qassim: well, 1) it's not good to throw exceptions in the controller, and 2) we need to add more proper handling for the exceptions

Al-Qassim: first, make a new directory called exceptions in the service directory: and inside it, define a custom exception as follows::
`class NotFoundException(message: String) : RuntimeException(message)`

second the function `getChatById` inside the `ChatService` should return a `Chat`, not a `Chat?`. if no chat exist with this id, the the `getChatById` inside the `ChatService` should throw `NotFoundException("no chat was found with id: $id")`

third: inside the `api/dto` directory, make a new class data class `ErrorResponse(val message: String)`, this will be return to the client as a body

forth: make an advisor class that will handle the errors automatically:
in the `api/controller` directory, make a new class as follows:

```
@RestControllerAdvice
@Order(1)
class ChatControllerAdvice : ResponseEntityExceptionHandler() {
    @ExceptionHandler(NotFoundException::class)
    fun handleNotFoundException(e: NotFoundException, request: WebRequest): ResponseEntity<ErrorResponse> {
        val error = ErrorResponse(e.message ?: "Resource not found")
        return ResponseEntity
            .status(HttpStatus.NOT_FOUND)
            .body(error)
    }
}
```

Al-Qassim: before: 500 ❌ | after: 404 ✅
--- | ---
image | image

Al-Qassim: for the `ChatController`, no exception should be thrown, just `val chat = chatService.getChatById(id)`

Al-Qassim (changes requested):
good job, but please apply the required notes

chat/src/main/kotlin/net/thechance/chat/api/controller/ErrorHandler.kt
Al-Qassim, line 1: rename the file to ChatControllerAdvice.kt

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
```diff
@@ -50,4 +56,24 @@ class ChatService(
...
+            imageUrl = otherUser?.imageUrl,
+            requesterId = userId,
+            id = chatId
+        )
+    }
+
+    private fun getChatName(contact: Contact?, user: ContactUser? ): String{
+        return contact?.let { "${it.firstName} ${it.lastName}" }
+            ?: user?.let { "${it.firstName} ${it.lastName}" }.orEmpty()
+    }
```
Al-Qassim, line 77: no need for all of this, I think the following is good enough
```
    fun getChatById(chatId: UUID, userId: UUID): ChatModel {
        val chat = chatRepository.findByIdOrNull(chatId) ?: throw NotFoundException("no chat was found with id: $chatId")
        val otherUser = chat.users.firstOrNull { it.id != userId }
        
        return ChatModel(
            name = getChatName(otherUser),
            imageUrl = otherUser?.imageUrl,
            requesterId = userId,
            id = chatId
        )
    }

    private fun getChatName(user: ContactUser?): String{
        return user?.let { "${it.firstName} ${it.lastName}" }.orEmpty()
    }
```

Al-Qassim (approved):

devfalah (approved):

dev-bilal-azzam (approved):

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
omer1998: Yes qassim, your approach is simpler but With this the name of the chat will be name of the mina user 
Not the name of this user as registered in contact list
So we decided to make the name of chat to be similar to the name of this user in the contact list of the user

PR #105, opened 2025-10-12
Feature/cht 57 picture storage

ahmedfikry24:
## what is the PR Scope ?
upload images on chat 

## why do we need that ?
let mena user able to send images on chat

 data class MessageResponseDto(
     val id: UUID,
     val senderId: UUID,
     val chatId: UUID,
     val text: String?,
     val images: List<String>,
     val sendAt: Instant,
     val isRead: Boolean
 )


chat/src/main/kotlin/net/thechance/chat/entity/MessageImages.kt
```diff
@@ -0,0 +1,18 @@
...
+data class MessageImages(
+    @Id @Column(columnDefinition = "uuid", updatable = false, nullable = false)
+    val id: UUID = UUID.randomUUID(),
+
+    @Column(nullable = false)
+    val url: String,
+
+    @ManyToOne(optional = false)
+    @JoinColumn(name = "message_id", referencedColumnName = "id", nullable = false)
+    val message: Message
```
Al-Qassim: no need for this `val message: Message`. I think you can remove it
you already stored a list of images inside the `Message` entity.
also, karrar have commented yesterday that this is not a good practice (storing both entities inside each other) it's not even good for the memory. unfortunately, we have this practice in a lot of places

chat/src/main/kotlin/net/thechance/chat/api/dto/MessageDto.kt
```diff
@@ -1,49 +1,40 @@
...
 import net.thechance.chat.entity.Message
-import net.thechance.chat.service.args.CreateMessageArgs
 import org.springframework.data.domain.Page
 import java.time.Instant
 import java.util.*
 
 data class MessageRequestDto(
     val chatId: UUID,
-    val text: String,
+    val messageId: UUID?,
```
Al-Qassim: why a `messageId` is in the `MessageRequestDto`?

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
```diff
@@ -31,23 +37,52 @@ class ChatService(
...
-        messageRepository.save(
-            Message(
-                id = message.id,
-                senderId = message.senderId,
-                chat = chat,
-                text = message.text,
-                sentAt = message.sendAt,
-            )
+    @Transactional
+    fun saveMessage(senderId: UUID, request: MessageRequestDto): Message {
```
Al-Qassim: in general, the architecture of the backend requires that the service layer doesn't know anything about the api layer, so we should not use the RequestDto or ResponceDto in the services

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
```diff
@@ -31,23 +37,52 @@ class ChatService(
...
+        request.messageId?.let { messageId ->
+            messageRepository.findById(messageId).orElse(null)?.let { return it }
+        }
```
Al-Qassim: I still not getting why we need a `messageId` inside the request?
what I know is that a message should not have any id if it's saved yet.
and during the saving process, the id will be generated for it. just like what the old code was doing

chat/src/main/kotlin/net/thechance/chat/service/AttachmentStorageService.kt
```diff
@@ -0,0 +1,63 @@
...
+
+@ConfigurationProperties(prefix = "storage.mena")
+data class ChatStorageProperties(
+    val bucket: String,
+    val cdnEndpoint: String
+)
+
+@Service
+@EnableConfigurationProperties(ChatStorageProperties::class)
+class AttachmentStorageService(
```
Al-Qassim, line 24: I think it will be better if we this class to just `StorageService`, instead of `AttachmentStorageService`.
because it can be used to store images for any usecase, not just attachments

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
Al-Qassim: also, why the function is returning a `Message`? 
intuitively, function with a `save` in its name is expected to return nothing, it's not a `get` function.

at most, it can return an id

chat/src/main/kotlin/net/thechance/chat/entity/MessageImages.kt
ahmedfikry24: Yea It's not required to put it here but I just thought that if need to get the message again but OK it's better if we remove it or replaced by message id

chat/src/main/kotlin/net/thechance/chat/service/AttachmentStorageService.kt
ahmedfikry24: Cause in chat feature any files call as chat attachments

chat/src/main/kotlin/net/thechance/chat/api/dto/MessageDto.kt
ahmedfikry24: MessageId require from client side so I add it but after changes I think we can remove it but still need to check it's works OK with mobile

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
ahmedfikry24: yea you are right i will change service funcs args, about fun return message i need to create message to use it

Al-Qassim (approved):

chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
```diff
@@ -25,11 +26,10 @@ class ChatController(
     @MessageMapping("/chat.privateMessage")
     fun sendPrivateMessage(@Payload chatMessage: MessageRequestDto, principal: Principal) {
```
omer1998: you can inject sender id directly here in this way 
fun sendPrivateMessage(@Payload chatMessage: MessageRequestDto, @AuthenticationPrincipal senderId: UUID) { ... }

chat/src/main/kotlin/net/thechance/chat/entity/Message.kt
```diff
@@ -16,13 +11,17 @@ data class Message(
...
     @Column(nullable = false)
     val isRead: Boolean = false,
 
+    @OneToMany(cascade = [CascadeType.ALL], orphanRemoval = true, fetch = FetchType.EAGER)
+    @JoinColumn(name = "message_id")
+    val images: List<MessageImages> = emptyList(),
+
     @ManyToOne(optional = false)
     @JoinColumn(name = "chat_id", referencedColumnName = "id", nullable = false)
     val chat: Chat
```
omer1998, line 28: i think the design of this table (messages) and relying on null value of text to determine the type of message is not perfect and not scalable in the future (but i think it is a good start), 
one possible approach is to define table for message metadata and table for each message type and these will be linked to the main message table.

chat/src/test/kotlin/net/thechance/chat/api/controller/ChatControllerTest.kt
```diff
@@ -52,15 +53,53 @@ class ChatControllerTest {
...
-
+        verify {
+            chatService.saveMessage(
+                match { it.senderId == senderId && it.text == "message1" && it.chatId == chatId },
+            )
+        }
     }
 
+    @Test
+    fun `sendMessageImages should upload image then send to user`() {
```
omer1998, line 73: it always better to make unit test for single specific thing

omer1998 (approved):
great work ahmed

chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
```diff
@@ -25,11 +26,10 @@ class ChatController(
...
         val senderId = UUID.fromString(principal.name)
-        val createdMessage = chatMessage.toCreateMessageArgs(senderId = senderId)
-        chatService.saveMessage(createdMessage)
+        val message = chatService.saveMessage(chatMessage.toReqArgs(senderId)).toDto()
```
Karrar-Mohammed: here we are doing two mapping toReqArgs then toDto, do we need that ? can we make it in one mapping ?

chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
```diff
@@ -49,12 +49,31 @@ class ChatController(
...
+        @PathVariable("chatId") chatId: UUID,
+        @RequestParam("images") images: List<MultipartFile>,
+        principal: Principal
+    ): ResponseEntity<MessageResponseDto> {
+        val senderId = UUID.fromString(principal.name)
+
+        val message = chatService.saveMessageImages(chatId, senderId, images).toDto()
+
+        sendMessageToUser(
+            user = chatId.toString(),
```
Karrar-Mohammed: here the parameter name is user but we are sending chatId, can we change the name to chatId ?

chat/src/main/kotlin/net/thechance/chat/entity/Message.kt
```diff
@@ -16,13 +11,17 @@ data class Message(
...
+    @Column(columnDefinition = "TEXT", nullable = true)
+    val text: String? = null,
     @Column(nullable = false)
     val sentAt: Instant = Instant.now(),
     @Column(nullable = false)
     val isRead: Boolean = false,
 
+    @OneToMany(cascade = [CascadeType.ALL], orphanRemoval = true, fetch = FetchType.EAGER)
+    @JoinColumn(name = "message_id")
+    val images: List<MessageImages> = emptyList(),
```
Karrar-Mohammed: i think we can simply save the image urls here, do we need to create entity for the images ?

chat/src/main/kotlin/net/thechance/chat/exception/ErrorCodes.kt
```diff
@@ -0,0 +1,6 @@
+package net.thechance.chat.exception
+
+object ErrorCodes {
+    const val INVALID_IMAGE_FORMAT = 1201
+    const val IMAGE_UPLOAD_FAILED = 1202
+}
```
Karrar-Mohammed, line 6: why do we need custom error codes ?

chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
ahmedfikry24: no unfortunutly, one for mapping args and another to map to dto, i update to be more readability

chat/src/main/kotlin/net/thechance/chat/entity/Message.kt
ahmedfikry24: i change it to create it with annotations only

chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
ahmedfikry24: Thanks omar for this info I will update it

chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
ahmedfikry24: It's websocket function parameter name called user so I just name it like socket

raouf-kamel (approved):

Karrar-Mohammed (approved):

PR #106, opened 2025-10-12
Feature: idn 101 prepare save user info

poula22:
## what is the PR Scope ?
update user info and upload new profile image to digital ocean

## why do we need that?
We need that to update user profile info

## update profile info Endpoint
POST `/identity/profile`

## Request
```json
{
    "username": "testuser51",
    "firstName": "Test12",
    "lastName": "User533",
    "birthDate": "1975-01-01",
    "gender": 1
}
```

## Response
```json
{
    "id": "68033e82-a1aa-4974-89c2-ce5435b652a9",
    "username": "testuser51",
    "firstName": "Test12",
    "lastName": "User533",
    "imageUrl": "storage.mena.cdn-endpoint/images/identity/profile/553af859-0688-433e-9148-d351ee74a451.jpg?time=2025-10-25T13:38:52.436341200",
    "birthDate": "1975-01-01",
    "gender": 1
}
```

## update user image Endpoint
POST `/identity/image`

## Request
image

## Response
```json 
{
    "username": "testuser5",
    "firstName": "Test",
    "lastName": "User5",
    "imageUrl": "E:/package/name/profile/testpng",
    "birthDate": "1970-01-01",
    "gender": 1
}
```

## Delete profile image Endpoint
DELETE `/identity/profile/image`

## Response
```json
{
    "message": "Image deleted successfully"
}
```


identity/src/main/kotlin/net/thechance/identity/entity/User.kt
```diff
@@ -1,27 +1,58 @@
 package net.thechance.identity.entity
 
+import com.fasterxml.jackson.annotation.JsonFormat
 import jakarta.persistence.Column
 import jakarta.persistence.Entity
 import jakarta.persistence.Id
 import jakarta.persistence.Table
+import jakarta.validation.constraints.Max
+import jakarta.validation.constraints.Min
```
Karrar-Mohammed: these are for the validation i don't think will do any effect on the database or the table, did you test it that it's working as expected ?

identity/src/main/kotlin/net/thechance/identity/api/controller/ProfileController.kt
```diff
@@ -14,23 +13,22 @@ import java.util.*
...
 class ProfileController(
-    private val userService: UserService,
-    private val updateUserProfileImageService: UpdateUserProfileImageService
+    private val userService: UserService
 ) {
 
+    @PostMapping("/me")
+    fun updateCurrentUserProfile(
+        @AuthenticationPrincipal userId: UUID,
+        @RequestPart("user") updateProfileRequest: UpdateProfileRequest,
```
Karrar-Mohammed: i noticed that you are using the validation annotation to validate the UpdateProfileRequest but for make these annotation work and do the validation you should here add the @Valid annotation

identity/src/main/kotlin/net/thechance/identity/api/dto/UpdateProfileRequest.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import jakarta.validation.constraints.Max
+import jakarta.validation.constraints.Min
+
+data class UpdateProfileRequest (
+    val username: String,
+    val firstName: String,
+    val lastName: String,
+    val imageUrl: String?,
+    val birthDate: String,
```
Karrar-Mohammed, line 15: if you need any of these field to be mandatory so you might need the not blank annotation

identity/src/main/kotlin/net/thechance/identity/entity/User.kt
```diff
@@ -1,27 +1,58 @@
...
+    val birthDate: LocalDate,
+
+    @field:Min(
+        value = Gender.MALE.toLong(),
+        message = "Gender must be ${Gender.MALE} (man) or ${Gender.FEMALE} (woman)"
+    )
+    @field:Max(
+        value = Gender.FEMALE.toLong(),
+        message = "Gender must be ${Gender.MALE} (man) or ${Gender.FEMALE} (woman)"
+    )
```
Karrar-Mohammed: i don't think these will work or add constraints in the table of the database

identity/src/main/kotlin/net/thechance/identity/mapper/util/dateFormmater.kt
```diff
@@ -0,0 +1,9 @@
+package net.thechance.identity.mapper.util
+
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
+
+const val DEFAULT_DATE_FORMAT = "yyyy-MM-dd"
+fun LocalDate.formatAsString(): String {
+    return format(DateTimeFormatter.ofPattern(DEFAULT_DATE_FORMAT))
+}
```
Karrar-Mohammed: why do we need that here in the backend ?

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -41,4 +45,25 @@ class UserService(
...
+            fileName = "$userId"
+        )
+
+        val updatedUser = user.copy(
+            username = updateProfileRequest.username,
+            firstName = updateProfileRequest.firstName,
+            lastName = updateProfileRequest.lastName,
+            imageUrl = imageUrl,
+            birthDate = LocalDate.parse(updateProfileRequest.birthDate),
+            gender = updateProfileRequest.gender,
```
Karrar-Mohammed: if you always update the user profile details then i think you should also validate the UpdateProfileRequest details

Karrar-Mohammed: also the UpdateProfileRequest is from the api package so we should not pass it to the service module because it's violating our architecture

identity/src/main/resources/db-identity/changelog/changelogs/changelogs_002.yaml
Karrar-Mohammed, line 1: we disabled the liquibase, so i don't think we need to write any migration so far

identity/src/main/kotlin/net/thechance/identity/api/dto/UpdateProfileRequest.kt
```diff
@@ -0,0 +1,37 @@
...
+    val firstName: String,
+    @field:NotBlank(message = "lastName must not be blank")
+    val lastName: String,
+    @field:NotBlank(message = "birthDate must not be blank")
+    val birthDate: String,
+    @field:Min(1, message = "Gender must be 1 or 2")
+    @field:Max(2, message = "Gender must be 1 or 2")
+    val gender: Int,
+    val imageUrl: String?,
+    val updateImage: Boolean = false
```
Karrar-Mohammed: why do we need this "updateImage" parameter ?

poula22: we have to update user info including image, and we have 3 scenarios:
- update info with image
- update info without image
- update info and delete image

for the last 2 options image will be null so I need a flag to tell function the action is delete image or ignore it

Karrar-Mohammed: why don't make the image deletion in separate end point ? because you need to delete the image from the s3 server, currently you just remove the url from the profile record in the database but the image is still exist in the server

AimanYosofiUlt: We currently follow an approach where both the user data and image are sent in a single request. so deleting the image should also be handled within the same request.

**If we decide to change the approach, it would require three separate endpoints (save user info, upload user image, and delete user image)**
However, we believe the current approach is better, as it ensures that all operations either succeed or fail together.

Karrar-Mohammed: but the with the current approach you need to send all the data even if you want to change one value, am i understanding the approach correctly ?

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -41,4 +45,30 @@ class UserService(
...
+            file != null -> identityImageStorageService.uploadImage(file = file, fileName = "${userEntity.id}")
+            else -> ""
+        }
+
+        val updatedUser = userEntity.copy(
+            username = user.username,
+            firstName = user.firstName,
+            lastName = user.lastName,
+            imageUrl = imageUrl,
+            birthDate = LocalDate.parse(user.birthDate),
```
Karrar-Mohammed: i suggest to do the parsing to local date in the controller in the mapper to the model

identity/src/main/kotlin/net/thechance/identity/api/utils/jsonBody.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.http.ResponseEntity
+
+fun <T> ResponseEntity.BodyBuilder.jsonBody(key: String, value: T): ResponseEntity<String> {
+    return contentType(APPLICATION_JSON)
+        .body(mapToJson(key, value))
+}
+
+private fun <T> mapToJson(key: String, value: T): String {
+    return "{\"$key\": \"$value\"}"
+}
```
Karrar-Mohammed: why do we need that ?

identity/src/main/kotlin/net/thechance/identity/service/IdentityImageStorageService.kt
```diff
@@ -26,17 +22,27 @@ class IdentityImageStorageService(
 ) {
     fun uploadImage(
         file: MultipartFile,
-        fileName: String,
-        folderName: String,
+        fileName: String = UUID.randomUUID().toString(),
```
Karrar-Mohammed: why are we putting random uuid as file name ? how will that benefit us ?

identity/src/main/kotlin/net/thechance/identity/api/controller/ProfileController.kt
```diff
@@ -1,36 +1,69 @@
...
         @AuthenticationPrincipal userId: UUID,
-        @RequestParam("file") file: MultipartFile,
-    ): ResponseEntity<GeneralResponse> {
-        updateUserProfileImageService(userId, file)
-        val response = GeneralResponse("Profile image uploaded successfully")
-        return ResponseEntity.ok(response)
+        @RequestPart("file") file: MultipartFile,
+    ): ResponseEntity<String> {
+        val imageUrl = userService.updateUserImage(userId, file)
+        return ResponseEntity.ok().jsonBody(IMAGE_URL_KEY, imageUrl)
```
Karrar-Mohammed: instead of using this jsonBody function, why not just simply create updateImageResponse and we return the image in this response dto

identity/src/main/kotlin/net/thechance/identity/mapper/UserMapper.kt
```diff
@@ -3,11 +3,14 @@ package net.thechance.identity.mapper
...
-		lastName = lastName,
-		profileImageUrl = imageUrl
-	)
+fun User.toResponse(cdnEndpoint: String): ProfileResponse {
+    return ProfileResponse(
+        id = id.toString(),
+        username = username,
+        firstName = firstName,
+        lastName = lastName,
+        imageUrl = if (imageUrl.isNullOrBlank()) null else "$cdnEndpoint/$imageUrl",
```
Karrar-Mohammed, line 12: why do we need that ? in my opinion if the imageUrl is null and it should not be null then we have a bug and we need to fix it, so i don't think we need to do this

poula22: The case here is I am saving image url as relative url in database, and when I send it to front end I need to add base url to file server. What should I do if there is no image url in database?

Without this check the result will be "baseUrl/null" but I want to be null

identity/src/main/kotlin/net/thechance/identity/api/utils/jsonBody.kt
poula22: I thought that I will dynamic response, but It's totally wrong as you mentioned before how can frontend know response then, so sorry for that🤦🏻‍♂️

identity/src/main/kotlin/net/thechance/identity/api/controller/ProfileController.kt
poula22: I will change that

identity/src/main/kotlin/net/thechance/identity/service/IdentityImageStorageService.kt
poula22: I forget to remove it by mistake there was 3 solution in my mind one of them is generating file name every time we need to upload image, but now we don't need it

AimanYosofiUlt (approved):

identity/src/main/kotlin/net/thechance/identity/mapper/UserMapper.kt
Karrar-Mohammed: ok i misunderstood this line, thanks for your clarification

identity/src/main/kotlin/net/thechance/identity/api/controller/ProfileController.kt
```diff
@@ -42,15 +43,17 @@ class ProfileController(
...
+        return ResponseEntity.ok(response)
     }
 
     @DeleteMapping("/image")
-    fun deleteUserImage(@AuthenticationPrincipal userId: UUID): ResponseEntity<String> {
+    fun deleteUserImage(@AuthenticationPrincipal userId: UUID): ResponseEntity<DeleteImageResponse> {
         userService.deleteUserImage(userId)
-        return ResponseEntity.ok().jsonBody(MESSAGE_KEY, "Image deleted successfully")
+        val response = DeleteImageResponse("Image deleted successfully")
+        return ResponseEntity.ok(response)
```
Karrar-Mohammed: i don't think this response will be benefit for the client, so if that the case then returning just ok without response body will be enough

Karrar-Mohammed (approved):
good job 👏

nabilBouzineDev (approved):

PR #107, opened 2025-10-13
Feature: TRN-56 update user categories

yusufnasserdev:
## What is the PR Scope?

This PR introduces a new PATCH endpoint for managing user categories, allowing partial updates to a user's selected categories. The implementation includes:

- New PATCH endpoint at `/categories` for adding/removing categories
- Request validation to prevent overlapping add/remove operations
- Comprehensive response with patch metadata and updated category list
- Service layer logic to handle partial category updates
- Proper error handling for non-existent users and categories

## Why do we need that?

A PUT endpoint would require sending the entire category selection, which is inefficient for partial updates. This PATCH endpoint provides:

- **Efficiency**: Only send the categories to add/remove instead of full list
- **Better UX**: Clients can make incremental changes without tracking full state
- **Performance**: Reduced payload size and database operations
- **Atomic operations**: Multiple add/remove operations in single request

## Endpoints with Request and Response Body:

**Endpoint:** `PATCH /trends/categories`

**Request Headers:**
- Authentication: Bearer token required

**Request Body:**
```json
{
  "add": [
    "6f37ddc8-2c06-4a03-af79-c5079dd4d61f",
    "0d733503-5447-495b-a1be-41c222817d25"
  ],
  "remove": [
    "c8aac40e-74ff-4240-b6eb-6bbac7bcc27c",
    "1c0e34d9-737d-4003-a9dc-2c7ad14a9ca6",
    "e878f644-564c-45f7-ac95-e728567a7962"
  ]
}
```

**Response Body:**
```json
{
  "patchMetadata": {
    "addedCount": 2,
    "removedCount": 3
  },
  "updatedCategories": [
    {
      "id": "c8aac40e-74ff-4240-b6eb-6bbac7bcc27c",
      "name": "Technology",
      "emoji": "💻",
      "isSelected": false
    },
    // ... full list of categories with updated selection status
  ]
}
```

### DIFF
--- a/trends/src/main/kotlin/net/thechance/trends/api/dto/PatchMetadata.kt
+++ b/trends/src/main/kotlin/net/thechance/trends/api/dto/PatchMetadata.kt
@@ -1,0 +1,6 @@
+package net.thechance.trends.api.dto
+
+data class PatchMetadata(
+    val addedCount: Int,
+    val removedCount: Int
+)
--- a/trends/src/main/kotlin/net/thechance/trends/api/controller/CategoryController.kt
+++ b/trends/src/main/kotlin/net/thechance/trends/api/controller/CategoryController.kt
@@ -1,10 +1,7 @@
 package net.thechance.trends.api.controller
 
 import jakarta.validation.Valid
-import net.thechance.trends.api.dto.category.CategoryResponse
-import net.thechance.trends.api.dto.category.SubmitUserCategoriesRequest
-import net.thechance.trends.api.dto.category.SubmitUserCategoriesResponse
-import net.thechance.trends.api.dto.category.toCategoryResponse
+import net.thechance.trends.api.dto.category.*
 import net.thechance.trends.service.CategoryService
 import net.thechance.trends.service.TrendUserService
 import org.springframework.http.ResponseEntity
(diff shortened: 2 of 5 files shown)


mohamedd-hassan (approved):

ahmeddsamii (commented):
I don't think we need to return the full list of reels, you can instead return only the updated ones

one more suggestion: i don't think we need to return the added count and removed but this is just a suggestion but it's okay if we did that

trends/src/main/kotlin/net/thechance/trends/api/dto/PatchMetadata.kt
```diff
@@ -0,0 +1,6 @@
+package net.thechance.trends.api.dto
+
+data class PatchMetadata(
```
ahmeddsamii, line 3: you can rename this to patchCategories because data is too generic

yusufnasserdev: I believe it should be generic as it can be used for other patch requests.

ahmeddsamii (approved):
Gread job 👏

devfalah (approved):

PR #111, opened 2025-10-16
refactor/CHT-74-let-backend-allow-subsecribtion-to-all-private-messages-at-once

Al-Qassim:
## what is the PR Scope ?
allow the client to subsection to all his private messages from all his chats at once, and add a new end point to return a chat summary by id.

there ware also some fixes and refactors. for example:
- putting extra necessary parameter in the MarkMassageAsReadEvent, like which message? and in which chat?
- removing some code to preserve the api/service/repo responsibility structure
- ...

## why do we need that ?

this is necessary to update the main chat screen in the mobile app, where it should listen to all the messages from all the user chats, and update the up based on new messages.
so no need for a new ws session each time the client need to get the messages of another chat.

end point: `/chat/chatsSummary/{chatId}`
response:
```
{
    "id": "ec5ffb28-308b-4f92-ae62-e3bfc319dc27",
    "name": "Nabil Bouzine",
    "imageUrl": "https://cdn.marketplaceevents.com/sitefinity/images/default-source/icons/sunflower.png",
    "lastMessage": {
        "text": "hello",
        "sentAt": "2025-10-16T14:20:58.141784Z",
        "isMine": true
    },
    "unReadMessagesCount": 1
}
```

### DIFF
--- a/chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
+++ b/chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
@@ -66,7 +66,7 @@ class ChatService(
         val lastMessages = messageRepository.findLastMessagesForChats(chatIds)
         val unreadCounts = chatRepository.findUnreadCountsForChats(chatIds)
 
-        val chatSummaries = chats.map { chat ->
+        val chatsSummaries = chats.map { chat ->
             val otherUser = chat.users.firstOrNull { it.id != userId }
             chat.toSummary(
                 userId,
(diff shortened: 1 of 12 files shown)


Karrar-Mohammed (approved):

raouf-kamel (approved):

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
```diff
@@ -39,17 +39,17 @@ class ChatService(
...
-            Message(
-                id = message.id,
-                senderId = message.senderId,
-                chat = chat,
-                text = message.text,
-                sentAt = message.sendAt,
-            )
+            message
         )
+        return message
```
dev-bilal-azzam, line 52: Why not `return messageRepository.save(message)`

dev-bilal-azzam (approved):

PR #114, opened 2025-10-22
Refactor/FTH-97-day prayer times

marwanqashwa:
## what is the PR Scope ?
Refactor prayer time data model and related logic 
Simplify DTOs
Use Instant for time
Switch remote client from WebClient to RestTemplate
Repackaging

## why do we need that ?
To make the codebase simpler, safer, and easier to maintain by removing unused fields, improving time handling

image


AhmadKharfan (approved):

Ahmed7osny1 (approved):

faith/src/main/kotlin/net/thechance/faith/api/controller/PrayerTimeController.kt
```diff
@@ -21,8 +23,18 @@ class PrayerTimeController(
...
+
+    private fun String.toLocalDate(): LocalDate = runCatching {
+        this.split('-').let {
+            val day = it[0].toIntOrNull().orZero()
+            val month = it[1].toIntOrNull().orZero()
+            val year = it[2].toIntOrNull().orZero()
+            LocalDate.of(year, month, day)
+
+        }
+    }.getOrDefault(LocalDate.of(1970, 1, 1))
```
Karrar-Mohammed: why do we need that? why not the client send the date as iso so you can just use the .toString function of the localDate

faith/src/main/kotlin/net/thechance/faith/entity/DayPrayerTimings.kt
```diff
@@ -3,68 +3,41 @@ package net.thechance.faith.entity
...
 import org.hibernate.annotations.CreationTimestamp
 import java.time.Instant
+import java.time.LocalDate
 
 @Entity
-@Table(name = "day_prayer_timings", schema = "faith")
+@Table(
+    name = "day_prayer_timings",
+    schema = "faith",
+    indexes = [Index(name = "idx_date_latitude_longitude", columnList = "date, latitude, longitude")]
```
Karrar-Mohammed: i suggest to not add any index currently, when we will face issue in the performance then we will add index based on our issue

faith/src/main/kotlin/net/thechance/faith/service/AyahBookmarkService.kt
```diff
@@ -13,14 +14,15 @@ class AyahBookmarkService(
...
     fun saveBookmark(ayahBookmark: AyahBookmark): AyahBookmark {
         return ayahBookmarkRepository.save(ayahBookmark)
     }
 
+    @Transactional
     fun deleteByIdAndUserId(id: Int, userId: UUID) {
-        if (ayahBookmarkRepository.existsByUserIdAndId(userId, id).not())
-            throw AyahBookmarkNotFoundException("Bookmark with id '$id' not found")
-        ayahBookmarkRepository.deleteById(id)
+        val exists = ayahBookmarkRepository.deleteByUserIdAndId(userId, id) > 0
```
Karrar-Mohammed: the name of the variable "exists" is not accurate because here you are deleting item so i suggest to name it "deleted" so it can be more accurate that you are using this variable to check if it item deleted or not

faith/src/main/resources/db-faith/changelog/changelogs/changelogs_002.yaml
Karrar-Mohammed, line 1: liquibase is not disabled do you think we need to write migration ?

faith/src/main/kotlin/net/thechance/faith/api/controller/PrayerTimeController.kt
```diff
@@ -3,11 +3,13 @@ package net.thechance.faith.api.controller
...
 import net.thechance.faith.service.PrayerService
+import net.thechance.faith.utils.orZero
 import org.springframework.http.ResponseEntity
 import org.springframework.web.bind.annotation.*
+import java.time.LocalDate
 
 @RestController
-@RequestMapping("faith/DayPrayerTimes")
+@RequestMapping("faith/prayer/times")
 class PrayerTimeController(
```
Karrar-Mohammed, line 13: i don't think the name "PrayerTime" is accurate because it means the times of the person that is praying so i think the name "PrayTimes" the times of the pray is more accurate

faith/src/main/kotlin/net/thechance/faith/api/controller/PrayerTimeController.kt
marwanqashwa: Yeah, I did it that way because of how we write dates on our side, but you’re right using ISO format with `LocalDate.parse()` would be much simpler.

faith/src/main/kotlin/net/thechance/faith/service/AyahBookmarkService.kt
marwanqashwa: Yeah, you’re right  “deleted” makes more sense here. I’ll update it.

faith/src/main/resources/db-faith/changelog/changelogs/changelogs_002.yaml
marwanqashwa: Yes, you’re right. I know Liquibase isn’t enabled. I added them because I don’t have access to the database, so it’ll be easier for the person who has access to add them.

faith/src/main/kotlin/net/thechance/faith/api/controller/PrayerTimeController.kt
marwanqashwa: I think “PrayerTimes” is actually the correct form “pray” is a verb , while “prayer” is a noun (the act of praying). So “PrayerTimes” literally means the times of the prayers, what do you think ?

Asiasama710 (approved):

faith/src/main/resources/db-faith/changelog/changelogs/changelogs_002.yaml
Karrar-Mohammed: i don't think it's needed we enabled the hibernate ddl, so it should be updated automatically, even if it for the person with access to the database then you can just send it to him, so either way we don't need to write liquibase migration

faith/src/main/kotlin/net/thechance/faith/api/controller/PrayerTimeController.kt
```diff
@@ -21,8 +22,12 @@ class PrayerTimeController(
...
             longitude = longitude,
-            date = date
+            date = date.toLocalDate()
         ).toResponse()
         return ResponseEntity.ok(prayerTimes)
     }
+
+    private fun String.toLocalDate(): LocalDate = runCatching {
+        LocalDate.parse(this)
+    }.getOrDefault(LocalDate.of(1970, 1, 1))
```
Karrar-Mohammed: why do we need a default value here ?

Karrar-Mohammed (approved):
good job 👏

faith/src/main/kotlin/net/thechance/faith/api/controller/PrayerTimeController.kt
marwanqashwa: I added the default value so that if the input date is invalid or can’t be parsed, it falls back to 1970-01-01, which is basically the start of the timestamp.
Do you think it would be better if I handle this case with a Bad Request instead of letting it continue and process the operation?

faith/src/main/resources/db-faith/changelog/changelogs/changelogs_002.yaml
marwanqashwa: You're right, I have access to the database now, so the Liquibase isn’t needed .

faith/src/main/kotlin/net/thechance/faith/api/controller/PrayerTimeController.kt
marwanqashwa: I changed it to handle the invalid date as a Bad Request instead of letting it continue and go through the whole process.

PR #115, opened 2025-10-23
refactor: restructure packages and organize classes

ElsayedMagdy122:
## what is the PR Scope ?
Refactor project structure and organize classes.

## why do we need that ?
The new structure separates concerns clearly

No new or modified endpoints.


mostafa-n3ma (approved):

dukan/src/main/kotlin/net/thechance/dukan/service/mapper/category/CategoryMapper.kt
```diff
@@ -1,6 +1,6 @@
-package net.thechance.dukan.mapper
+package net.thechance.dukan.service.mapper.category
```
TarekIdrees: Since we want to keep the service layer isolated from DTO concerns, it makes more sense to place the mapper package within the api package, where the actual mapping between entities and DTOs occurs.

ElsayedMagdy122: You're right — that's a good point

ElsayedMagdy122: I will modify it

haideraljawasim (approved):

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -3,9 +3,9 @@ package net.thechance.dukan.service
 import jakarta.persistence.EntityNotFoundException
 import jakarta.transaction.Transactional
 import net.thechance.dukan.entity.DukanProduct
-import net.thechance.dukan.exception.DukanProductCreationFailedException
-import net.thechance.dukan.exception.ProductNameAlreadyTakenException
-import net.thechance.dukan.exception.ProductNotFoundException
+import net.thechance.dukan.api.exception.DukanProductCreationFailedException
+import net.thechance.dukan.api.exception.ProductNameAlreadyTakenException
+import net.thechance.dukan.api.exception.ProductNotFoundException
```
Karrar-Mohammed: exceptions should not be in the api module because we use it in the services, as you see here the service now access something from the api package

dukan/src/main/kotlin/net/thechance/dukan/api/constants/EndPoints.kt
```diff
@@ -0,0 +1,5 @@
+package net.thechance.dukan.api.constants
```
Karrar-Mohammed: i don't see the name of this package "constants" is meaningful i suggest to call it util or something like that

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
ElsayedMagdy122: You're right

TarekIdrees (approved):

Karrar-Mohammed (approved):

PR #117, opened 2025-10-24
Refactor/cht 78 chat service controller

omer1998:
## what is the PR Scope ?
refactor chat services/ controller

## why do we need that ?
in order to organize the code structure and improve readability

## end points changes:
/{chatId}/summary --> for chat summary
/{chatId}/messages --> for messages of certain chat

make the error response in this way 
data class ErrorResponse(
    val message : String,
    val code: Int? = null
)
to generalize the response structure in case of failure and add optional error code related to image loading functionality


chat/src/main/kotlin/net/thechance/chat/entity/Message.kt
```diff
@@ -23,7 +23,7 @@ data class Message(
     @Column(name = "url", nullable = false)
     val images: List<String> = emptyList(),
 
-    @ManyToOne(optional = false)
+    @ManyToOne(optional = false, fetch = FetchType.LAZY)
     @JoinColumn(name = "chat_id", referencedColumnName = "id", nullable = false)
     val chat: Chat
```
Karrar-Mohammed: i think changing this to chatId of type UUID will be enough here

chat/src/main/kotlin/net/thechance/chat/service/AttachmentStorageService.kt
```diff
@@ -1,7 +1,7 @@
 package net.thechance.chat.service
 
-import net.thechance.chat.exception.ImageUploadFailedException
-import net.thechance.chat.exception.InvalidImageFormatException
+import net.thechance.chat.api.exception.ImageUploadFailedException
+import net.thechance.chat.api.exception.InvalidImageFormatException
```
Karrar-Mohammed: i think you moved the exception to the api module and now the service is access something from the api module, so we should fix that

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
```diff
@@ -26,16 +29,29 @@ class ChatService(
...
         val users = setOf(userId, receiverId)
-
-        val existingChat = chatRepository.findByUsersIds(users)
-        if (existingChat != null) return existingChat
-
         val requester = contactUserService.getUserById(userId)
         val otherUser = contactUserService.getUserById(receiverId)
+        val contact = otherUser.let {
+            contactService.getContactByOwnerIdAndContactUserId(userId, it.id)
+        }
```
Karrar-Mohammed: i don't think you need to use let here

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
```diff
@@ -26,16 +29,29 @@ class ChatService(
...
+
+        val newChat = chatRepository.save(Chat(users = mutableSetOf(requester, otherUser))).toModel(
+            chatName = getChatName(contact, otherUser),
+            imageUrl = otherUser.imageUrl.orEmpty(),
+            requesterId = userId
+        )
 
-        return chatRepository.save(Chat(users = mutableSetOf(requester, otherUser)))
+        return newChat
```
Karrar-Mohammed: this function became very big, can we split it into multiple smaller functions?

chat/src/main/kotlin/net/thechance/chat/entity/Message.kt
omer1998: yes karar this has been solved in another pr

Karrar-Mohammed (approved):

dev-bilal-azzam (approved):

chat/src/main/kotlin/net/thechance/chat/api/dto/ChatResponse.kt
```diff
@@ -1,10 +1,7 @@
 package net.thechance.chat.api.dto
 
-import net.thechance.chat.entity.Chat
-import net.thechance.chat.entity.Contact
-import net.thechance.chat.entity.ContactUser
 import net.thechance.chat.service.model.ChatModel
-import java.util.UUID
+import java.util.*
```
Al-Qassim, line 4: try not to use the `*` in the import, just import what you need

chat/src/main/kotlin/net/thechance/chat/api/dto/ChatSummaryResponse.kt
```diff
@@ -1,6 +1,6 @@
 package net.thechance.chat.api.dto
 
-import net.thechance.chat.entity.ChatSummary
+import net.thechance.chat.service.model.ChatSummary
 import org.springframework.data.domain.Page
 import java.time.Instant
 import java.util.*
```
Al-Qassim: same

Al-Qassim (approved):

raouf-kamel (approved):

PR #118, opened 2025-10-24
Refactor images to messages relation one image per message

dev-bilal-azzam:
## what is the PR Scope ?
- refactor Image - `Message` entity relation
## why do we need that ?
- structure each image in a `Message`
/chat/image

request body :
```
data class MessageImageRequest(
    val chatId: UUID,
    val image: MultipartFile
)
```

response  :
```
data class MessageResponse(
    val id: UUID,
    val senderId: UUID,
    val chatId: UUID,
    val text: String?,
    val imageUrl: String?,
    val sendAt: Instant,
    val isRead: Boolean,
    val isMine: Boolean
)
```

### DIFF
--- a/chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
+++ b/chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
@@ -1,6 +1,5 @@
 package net.thechance.chat.service
 
-import jakarta.persistence.EntityManager
 import net.thechance.chat.entity.*
 import net.thechance.chat.repository.ChatRepository
 import net.thechance.chat.repository.MessageRepository
@@ -22,8 +21,7 @@ class ChatService(
     private val chatRepository: ChatRepository,
     private val contactUserService: ContactUserService,
     private val attachmentStorageService: AttachmentStorageService,
-    private val contactService: ContactService,
-    private val entityManager: EntityManager
+    private val contactService: ContactService
 ) {
     @Transactional
     fun getOrCreateConversationByParticipants(userId: UUID, receiverId: UUID): Chat {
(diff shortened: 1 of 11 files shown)


raouf-kamel (approved):
Good job

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
```diff
@@ -40,35 +38,34 @@ class ChatService(
...
-        return messageRepository.save(message)
     }
 
     @Transactional
     fun saveMessageImage(args: MessageImageRequestArgs): Message {
-        val message = saveMessage(MessageRequestArgs(args.chatId, args.senderId, null, args.messageId))
         val imageUrl = attachmentStorageService.uploadImage(
             file = args.image,
-            fileName = args.image.originalFilename ?: "${message.id}-Untitled",
+            fileName = args.image.originalFilename ?: "${UUID.randomUUID()}-Untitled",
```
Karrar-Mohammed: why are we adding random uuid to the name ? it will not add any value actually it may mislead us, so i suggest  to either just delete it or replace it with the timestamp the current time so we can know when this image has been saved

Karrar-Mohammed (commented):
good job, let's just fix this simple comment before merging the PR

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
dev-bilal-azzam: done

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
```diff
@@ -40,35 +38,34 @@ class ChatService(
...
             folderName = FOLDER_NAME
         )
-        val updatedMessage = message.copy(
-            images = message.images.toMutableList().apply {
-                add(imageUrl)
-            }
+
+        return messageRepository.save(
+            Message(
+                id = UUID.randomUUID(),
```
Karrar-Mohammed: the Message id already has default value which is "UUID.randomUUID()" so no need to set it every time

Karrar-Mohammed (approved):

Al-Qassim (approved):

omer1998 (approved):

PR #120, opened 2025-10-25
Refactor: IDN-120 refactor address service

ThoriaHamdy99:
## what is the PR Scope ?
- Refactor address service

## why do we need that ?
- To make the code cleaner

- No new end points


identity/src/main/kotlin/net/thechance/identity/api/controller/AddressControllerAdvice.kt
```diff
@@ -19,14 +19,14 @@ class AddressControllerAdvice {
...
-    fun handleValidationExceptions(exception: MethodArgumentNotValidException): ResponseEntity<Map<String, String?>> {
+    fun handleValidationExceptions(exception: MethodArgumentNotValidException): ResponseEntity<ErrorResponse?> {
         val errors = exception.bindingResult.fieldErrors.associate {
             it.field to it.defaultMessage
         }
         logger.error("Validation failed: $errors", exception)
         return ResponseEntity
             .status(HttpStatus.BAD_REQUEST)
-            .body(errors)
+            .body(ErrorResponse("Data not valid"))
```
poula22: for better code style move this to variable

identity/src/main/kotlin/net/thechance/identity/api/controller/AddressControllerAdvice.kt
```diff
@@ -84,4 +60,12 @@ class AddressControllerAdvice {
             .status(HttpStatus.FORBIDDEN)
             .body(ErrorResponse(exception.message ?: "Address Can Not Be Updated"))
     }
+
+    @ExceptionHandler(DataNotValidException::class)
+    fun handleDataNotValidException(exception: DataNotValidException): ResponseEntity<ErrorResponse?> {
+        logger.error(exception.message)
+        return ResponseEntity
+            .status(HttpStatus.BAD_REQUEST)
+            .body(ErrorResponse(exception.message ?: "Data not valid"))
```
poula22: it's better to have variable for ErrorResponse(exception.message ?: "Data not valid") to be more readable

identity/src/main/kotlin/net/thechance/identity/api/controller/IdentityControllerAdvice.kt
```diff
@@ -19,14 +19,14 @@ class IdentityControllerAdvice {
...
-    fun handleValidationExceptions(exception: MethodArgumentNotValidException): ResponseEntity<Map<String, String?>> {
+    fun handleValidationExceptions(exception: MethodArgumentNotValidException): ResponseEntity<ErrorResponse?> {
         val errors = exception.bindingResult.fieldErrors.associate {
             it.field to it.defaultMessage
         }
         logger.error("Validation failed: $errors", exception)
         return ResponseEntity
             .status(HttpStatus.BAD_REQUEST)
-            .body(errors)
+            .body(ErrorResponse("Data not valid"))
```
poula22: It's better to make ErrorResponse("Data not valid") in a variable

identity/src/main/kotlin/net/thechance/identity/service/AddressService.kt
```diff
@@ -1,52 +1,36 @@
...
-        val user = userService.findById(userId)
-        return try {
-            val address = if (addressRepository.findAll().isEmpty()) createAddress(user, addressToAdd, true)
-            else createAddress(user, addressToAdd)
-            addressRepository.save(address)
-        } catch (_: Exception) {
-            throw AddressNotAddedException()
-        }
+    fun addAddress(userId: UUID, addressToAdd: AddressModel): Address {
+        val isThereAnAddress = addressRepository.existsByUserId(userId)
```
poula22, line 21: Why do we need this check here?

poula22 (approved):
great job👏
just some comments for better code style

identity/src/main/kotlin/net/thechance/identity/api/controller/AddressControllerAdvice.kt
```diff
@@ -19,14 +19,14 @@ class AddressControllerAdvice {
 
     @ExceptionHandler(MethodArgumentNotValidException::class)
     @ResponseStatus(HttpStatus.BAD_REQUEST)
-    fun handleValidationExceptions(exception: MethodArgumentNotValidException): ResponseEntity<Map<String, String?>> {
+    fun handleValidationExceptions(exception: MethodArgumentNotValidException): ResponseEntity<ErrorResponse?> {
```
Karrar-Mohammed: i don't think we need to make the ErrorResponse nullable because as i see we are always returning error

identity/src/main/kotlin/net/thechance/identity/api/controller/AddressControllerAdvice.kt
```diff
@@ -84,4 +60,12 @@ class AddressControllerAdvice {
             .status(HttpStatus.FORBIDDEN)
             .body(ErrorResponse(exception.message ?: "Address Can Not Be Updated"))
     }
+
+    @ExceptionHandler(DataNotValidException::class)
+    fun handleDataNotValidException(exception: DataNotValidException): ResponseEntity<ErrorResponse?> {
```
Karrar-Mohammed: same here

Karrar-Mohammed: also i see that in many places if it possible fix it in all places

identity/src/main/kotlin/net/thechance/identity/repository/AddressRepository.kt
```diff
@@ -2,12 +2,21 @@ package net.thechance.identity.repository
...
 import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.data.jpa.repository.Modifying
+import org.springframework.data.jpa.repository.Query
+import org.springframework.data.repository.query.Param
+import org.springframework.transaction.annotation.Transactional
 import java.util.*
 
 interface AddressRepository : JpaRepository<Address, UUID> {
     fun findByIdAndUserId(id: UUID, userId: UUID): Address?
+    fun findByUserIdOrderByCreatedAtAsc(userId: UUID): List<Address>
```
Karrar-Mohammed: i prefer to use pagination for the addresses because as far as i know we don't limit the user for creating addresses, and if we don't need in the mobile then in the mobile we can always ask for the first page, but we should make it pagination in the BE to be safe

Karrar-Mohammed (commented):
good job, the code now looks better, but let's fix these comments before merging the PR

identity/src/main/kotlin/net/thechance/identity/service/AddressService.kt
ThoriaHamdy99: because if there is an address so there is an active address if not so i will make the address as an active before adding it

identity/src/main/kotlin/net/thechance/identity/repository/AddressRepository.kt
ThoriaHamdy99: Actually, you are right from technical side as the user can enter a lot of addresses but in reality, the user can only have a 20 address at most so I need to check if we can limit them or not. if not so we will make a pagination.
But for now, the change will require changes in mobile, and we have to focus on sprint user stories so can we postpone this change?

Karrar-Mohammed: actually it's not a big just all it need is just a couple of minutes

Karrar-Mohammed (approved):

identity/src/main/kotlin/net/thechance/identity/repository/AddressRepository.kt
ThoriaHamdy99: Okay I made it from my side Just I will push it after testing and will add a default value for a page and page size to not enforce the mobile to change the code.

AhmadKharfan (approved):

PR #121, opened 2025-10-25
Refactor/trn 79 refactor backend

mohamedd-hassan:
## what is the PR Scope ?
Refactor trends module based on review meeting feedback

## why do we need that ?
to increase performance and improve code readability

same response dto but pagination starts at 0 instead of 1

DELETE `/trends/{trendId}/like` NEW ENDPOINT TO UNLIKE

`trends/user/reels` -> `trends/user`
`trends/user/categories/status` -> DELETED THE ENDPOINT

`/trends/reels/feed` & `trends/reels/feed/{id}` -> `/trends/feed` & `trends/feed/{id}`
DELETE `/trends/reels/{id}` -> `/trends/{id}` 
POST `/trends/reels` -> `/trends`
PUT `/trends/reels/{id}` -> PATCH `/trends/{id}`
PUT `/trends/reels/thumbnail/{reelId}`  -> PATCH `/trends/{trendId}/thumbnail`
POST `/trends/reels/view/{reelId}` -> `/trends/{trendId}/view` 
POST `/trends/reels/like/{reelId}` -> BEHAVIOUR CHANGE IT DOES NOT TOGGLE ANYMORE ONLY LIKES + `/trends/{trendId}/like`


trends/src/main/kotlin/net/thechance/trends/models/TrendWithLikeStatus.kt
```diff
@@ -0,0 +1,8 @@
+package net.thechance.trends.models
+
+import net.thechance.trends.entity.Trend
+
+interface TrendWithLikeStatus {
```
devfalah, line 5: why not normal class ?

trends/src/main/kotlin/net/thechance/trends/models/TrendUrls.kt
```diff
@@ -0,0 +1,6 @@
+package net.thechance.trends.models
+
+interface TrendUrls {
```
devfalah, line 3: same

trends/src/main/kotlin/net/thechance/trends/models/TrendWithLikeStatus.kt
mohamedd-hassan: while looking at spring boot documentation I found the projections feature for the repository needs an interface, I'm not sure if its possible to do so in kotlin with a normal data class

devfalah (approved):
well done

iZyadNasser (approved):
LGTM

ahmeddsamii (approved):

PR #123, opened 2025-10-27
create Rename Shelf  endpoint

mostafa-n3ma:
## what is the PR Scope ?
This PR adds the ability to rename (update) an existing shelf for a Dukan.

## why do we need that ?
Currently, shelves can only be created and deleted, but not renamed.
This update allows users to update a shelf title while maintaining validation rules and ownership checks.

Method: PUT
Path: /dukan/shelf/{shelfId}
request body:
{
  "title": "New Shelf Title"
}

response:
HTTP 204 No Content


dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanShelfController.kt
```diff
@@ -33,6 +33,16 @@ class DukanShelfController(
         return ResponseEntity.ok().build()
     }
 
+    @PutMapping("/{shelfId}")
+    fun renameShelf(
```
devfalah: Let's rename it to `updateShelf` in case we add more properties to the shelf in the future

dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
```diff
@@ -32,6 +32,21 @@ class DukanShelfService(
         )
     }
 
+    fun renameShelf(ownerId: UUID, shelfId: UUID, newTitle: String) {
```
devfalah: same

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanShelfController.kt
mostafa-n3ma: that's fair enough 👌🏻, done

dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
```diff
@@ -32,6 +32,21 @@ class DukanShelfService(
...
+        val dukan = dukanService.getDukanByOwnerId(ownerId)
+
+        if (dukanShelfRepository.existsByTitleAndDukanId(newTitle, dukan.id)) {
+            throw ShelfNameAlreadyTakenException()
+        }
```
devfalah: I think you can reduce the number of queries by removing this check. 
If there’s a unique constraint on the title column and you try to add the same title again the save method will throw an exception you can map it to ShelfNameAlreadyTakenException

mostafa-n3ma: I will check that

mostafa-n3ma: Good catch!
I agree — it’s better to remove the manual existsByTitleAndDukanId check to reduce the number of queries and rely on the database constraint instead.

This approach will correctly handle both of these cases:
	•	Creating a new shelf with a title that already exists → the DB unique constraint (dukan_id, title) will throw an exception.
	•	Updating (renaming) a shelf to another shelf’s existing title → also triggers the same constraint violation and returns the proper error.

However, when updating a shelf with the same title it already has, that’s not considered a database constraint violation, because it’s still the same row with the same unique key — so the update succeeds and responds with 204 No Content.

Example:
	•	Shelf A → "Electronics"
	•	Shelf B → "Clothes"

If I rename Shelf B to "Electronics", the DB constraint fires and prevents the duplicate ✅
But if I rename Shelf A to "Electronics" again (same name as before), there’s no conflict, so the request succeeds with 204 No Content ✅

To make this clearer at the domain level, I plan to add a custom NoChangesDetectedException in the update function.
This will explicitly handle the “same title” case and provide a clearer API response.

What do you think about this approach?

@devfalah

haideraljawasim: I think we need this check here, cause we need to map the exception to something we know to handle it in mobile rather than depending what database throw

dukan/src/main/kotlin/net/thechance/dukan/api/dto/shelf/DukanShelfTitleRequest.kt
```diff
@@ -3,7 +3,7 @@ package net.thechance.dukan.api.dto.shelf
 import jakarta.validation.constraints.NotBlank
 import jakarta.validation.constraints.Pattern
 import jakarta.validation.constraints.Size
-data class DukanShelfCreationRequest(
+data class DukanShelfTitleRequest(
```
haideraljawasim, line 6: i would prefer them separated like DukanShelfCreationRequest and DukanShelfUpdateRequest data class using the same will make them more coupled

dukan/src/main/kotlin/net/thechance/dukan/service/DukanShelfService.kt
mostafa-n3ma: we will map it to dukan exception just with calling db query relaying on the db constrains exceptions @haideraljawasim

dukan/src/main/kotlin/net/thechance/dukan/api/dto/shelf/DukanShelfTitleRequest.kt
mostafa-n3ma: you only have the title now sent when creating and when updating !!!!

AbdelrhmanAyman20470 (approved):

devfalah (approved):

ElsayedMagdy122 (approved):

PR #124, opened 2025-10-27
Feature/dkn 104 edit product

haideraljawasim:
## what is the PR Scope ?
- Implement endpoint and logic for update product 

## why do we need that ?
- we needed the endpoint to existing product  

Put request 
`dukan/product/{productId}`

request body 
```
{
  "name": "Organic Honey 500g",
  "description": "This is pure organic honey collected from natural hives. It has no added sugar or preservatives, ensuring 100% authenticity and freshness for your health and taste. Perfect for tea, baking, or as a natural sweetener in any meal.",
  "price": 12.99,
  "shelfId": "323b4008-b912-46d1-a2b5-5535139896be",
  "imageUrls": [
    "https://example.com/images/honey1.jpg",
    "https://example.com/images/honey2.jpg"
  ]
}
```

response :
```
{
  "productId": "9f77e6ad-2fd4-4e14-bf43-42479ddcc594"
}

```

also we implement an endpoint to upload single image of product 

`dukan/product/{productId}/image
`
request param :
file :MultipartFile


dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanProductController.kt
```diff
@@ -60,4 +63,26 @@ class DukanProductController(
...
+    ): DukanProductUpdateResponse {
+        val productUpdateParams = request.toProductUpdateParams(userId)
+        val productId = dukanProductService.updateProduct(productId, productUpdateParams)
+
+        return DukanProductUpdateResponse(productId)
+    }
+
+    @PostMapping("/{productId}/image")
+    fun uploadProductImage(
+        @AuthenticationPrincipal ownerId: UUID,
```
TarekIdrees: Let's stick with userId at the API level

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -83,6 +85,68 @@ class DukanProductService(
...
+        params: DukanProductUpdateParams
+    ): UUID {
+        val product = dukanProductRepository
+            .findByIdAndDukan_OwnerId(productId, params.ownerId)
+            .orElseThrow {
+                ProductNotFoundException()
+            }
+
+        if (product.name != params.name) {
+            existingProductByDukanIdAndName(product.dukan.id, params.name)
```
TarekIdrees: It would be clear if we choose a more expressive name like (checkProductNameExistance)

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -83,6 +85,68 @@ class DukanProductService(
...
+        }
+
+        val shelf = dukanShelfRepository.getReferenceById(params.shelfId)
+
+        deleteProductImages(product.imageUrls, params.imageUrls)
```
TarekIdrees: Since we have decided before to divide the creation of the product and uploading the images, I would prefer to do the same here.

example: 
```
val updatedProductId = product.copy(
            name = params.name.trim(),
            price = params.price,
            imageUrls = **_product.imgeUrls_**
            description = params.description.trim(),
            shelf = shelf
        ).also { dukanProductRepository.save(it) }.id
```
There will be a separate function called updateProductImages, which takes a list of files and includes logic to check whether this list contains any elements that differ from the existing product images.

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -83,6 +85,68 @@ class DukanProductService(
...
+            price = params.price,
+            imageUrls = params.imageUrls,
+            description = params.description.trim(),
+            shelf = shelf
+        ).also { dukanProductRepository.save(it) }.id
+
+        return updatedProductId
+    }
+
+    fun uploadProductImage(ownerId: UUID, productId: UUID, file: MultipartFile): String {
```
TarekIdrees: We don't need the ownerid where `**_dukanProductRepository.findById(productId)_**` will do the job, right?

TarekIdrees: Also, we can update this function to do the mentioned idea of updating the product images

TarekIdrees (changes requested):
If you want we can join a meeting t discuss in more details.

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanProductController.kt
haideraljawasim: you're right done

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
haideraljawasim: we needed the owner id to ensure that logged in user can only update products from its dukan

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
haideraljawasim: I didn't get you here , do you mean this updateProductImages function take multipart file or json file , if mulitpart file how could i track what the files where deleted and if only the urls images i don't see it necessary  as it can be done with single request

haideraljawasim:
> If you want we can join a meeting t discuss in more details.

Great , we can discuss this more in our daily meeting

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanProductRepository.kt
```diff
@@ -5,11 +5,13 @@ import org.springframework.data.domain.Page
...
 import org.springframework.data.domain.Pageable
+import org.springframework.data.jpa.repository.Query
 import java.util.*
 
 @Repository
 interface DukanProductRepository : JpaRepository<DukanProduct, UUID> {
     fun existsByDukanIdAndNameIgnoreCase(dukanId: UUID, name: String): Boolean
     fun findAllByShelfId(shelfId: UUID, pageable: Pageable): Page<DukanProduct>
     fun existsByShelfId(shelfId: UUID): Boolean
+    fun findByIdAndDukan_OwnerId(id: UUID, ownerId: UUID): Optional<DukanProduct>
```
Karrar-Mohammed: i don't we need the '_' in the name of this function

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -83,6 +82,72 @@ class DukanProductService(
         }
     }
 
+    @Transactional
+    fun updateProduct(
+        productId: UUID,
+        updateParams: DukanProductUpdateParams
```
Karrar-Mohammed: i think we can add the productId to the DukanProductUpdateParams, since it's for the update params and we need the productId for the update functionality

dukan/src/main/kotlin/net/thechance/dukan/api/dto/product/DukanProductUpdateRequest.kt
```diff
@@ -0,0 +1,22 @@
...
+    @field:Size(min = 100, max = 3000, message = "description must be between 100 and 3000 characters")
+    @field:NotBlank(message = "description must not be blank")
+    val description: String,
+    @field:NotNull(message = "price is required")
+    @field:DecimalMin(value = "0.001", inclusive = true, message = "price must be >= 0.001")
+    val price: Double,
+    @field:NotNull(message = "must choose at least one shelf")
+    val shelfId: UUID,
+    @field:Size(min = 1, max = 10, message = "must have at least one image and not execute then 10")
+    val imageUrls: List<String>
```
Karrar-Mohammed, line 21: i suggest to edit the product images in a separate end point, because this is make the request complicated and if one of the images failed to delete or to upload that will add logic to handle these cases, and we might need to upload the images again which sometimes unnecessary

haideraljawasim: hmmm , actually  i didn't get you here in the update procces we upload the images alone and we only append the urls to data base here ,so what i mean we have upload image endpoint that take one file and return url to client side then in client side we append these urls inside this request  body

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanProductRepository.kt
haideraljawasim: you're right at first i don't know why but  the ORM couldn’t generate the query , so i used  '_'

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
haideraljawasim: done

dukan/src/main/kotlin/net/thechance/dukan/api/dto/product/DukanProductUpdateRequest.kt
haideraljawasim: also for delete we thinking about create like table and add failed urls to, and have scheduler to try delete them later

TarekIdrees (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/dto/product/DukanProductUpdateRequest.kt
Karrar-Mohammed: ok maybe i misunderstood the  implementation, thanks for the clarification

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -83,6 +82,71 @@ class DukanProductService(
...
+            oldImageUrls
+                .filterNot { it in updatedImageUrls }
+                .forEach {
+                    if (imageStorageService.deleteImage(it).not()) {
+                        // TODO save failed images table and try to delete them later
+                    }
+                }
+        } catch (_: Exception) {
+            // TODO save failed images table and try to delete them later
+        }
```
Karrar-Mohammed, line 146: if you are planning to do that then maybe we can create something like queue to delete images so we don't even block the request for deleting the images we can simply just add it to the queue and return success then there will be a process to take from the queue and delete the images and if one of the images failed then we will just add it again to the queue,

try to research on how can we implement that and contact me if you need help, but let's do that in a separate PR

Karrar-Mohammed (approved):

ElsayedMagdy122 (approved):
Great Work Haider 🤎

PR #125, opened 2025-10-28
Feature/cht 81 create delete chat endpoint

omer1998:
## what is the PR Scope ?
create delete chat endpoint

## why do we need that ?
to enable user to delete specific chat

endpoint: /chat/delete/{chatId}
response: 
{
    "success": true,
    "message": "Successfully deleted chat 8c691d3b-109a-4287-a9eb-87ff3e1fe0fa"
}


Karrar-Mohammed:
i think this PR need to fix it's base or to update it with the latest develop version because i see many changes that is not related to the PR title

chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
```diff
@@ -129,6 +129,24 @@ class ChatController(
...
+    @DeleteMapping("/delete/{chatId}")
+    fun deleteChatById(
+        @PathVariable chatId: UUID
+    ): ResponseEntity<Unit> {
+        chatService.deleteChatById(chatId)
+        sendEventToUsers(chatId, DeleteChatResponse(chatId))
+        return ResponseEntity.ok().body(Unit)
+    }
+
+    private fun sendEventToUsers(chatId: UUID, eventPayload: Any) {
```
Karrar-Mohammed: use the name event here in my opinion is bit misleading or not very accureate, i prefer to call it sendMessageToUsers or notifyUsers it will be better, because for the first time when i read the name of the function i thought we are sending a domain event here, so it will make issue because we are using the same word for the event bus

chat/src/main/kotlin/net/thechance/chat/entity/Chat.kt
```diff
@@ -17,4 +17,19 @@ data class Chat(
...
+        this.cleanUpStatus = cleanUpStatus
+        return this
+    }
+}
+
+enum class CleanUpStatus{
+    NONE,
+    S3_DELETED_FAILED,
+    CLEANUP_FAILED,
+}
```
Karrar-Mohammed: why do we need that ? and why it's here ?

omer1998: this is to track the status of deleting specific chat
deleting chat involve deleting messages and the media associated with messages (like images)
if any problem happens during any step of these, the scheduler will run specifically on that chat and retry the deleting process

chat/src/main/kotlin/net/thechance/chat/api/controller/ChatController.kt
```diff
@@ -129,6 +129,24 @@ class ChatController(
         return ResponseEntity.ok(chat.toResponse())
     }
 
+    @DeleteMapping("/delete/{chatId}")
```
Al-Qassim: just `@DeleteMapping("/{chatId}")`, no need for `@DeleteMapping("/delete/{chatId}")`

chat/src/main/kotlin/net/thechance/chat/api/dto/DeleteChatResponse.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.chat.api.dto
+
+import java.util.UUID
+
+data class DeleteChatResponse (
+    val deletedChatId: UUID
+)
```
Al-Qassim, line 7: no need for a response

Al-Qassim: you can delete this file

chat/src/main/kotlin/net/thechance/chat/entity/Chat.kt
Al-Qassim: @omer1998 
can't you setup something like a garbage collector?, where you don't add a new column to the entity, just go to the messages and media and see if they corresponds to a deleted chat, if yes, delete them

Karrar-Mohammed: alqassim idea is good, i prefer at least to create a different table for that, because adding column to the chat table i don't see it reasonable

chat/src/main/kotlin/net/thechance/chat/api/dto/DeleteChatResponse.kt
Karrar-Mohammed: i thought of that also but it might need it so the other people also get notified to delete the chat from the other side i'm not sure if this the intention of this response we should wait for omer response to explain why we need that

omer1998: yes exactly, as mentioned by karrar, when deleting chat, we send this event to both users that this chat was deleted so navigate back.

chat/src/main/kotlin/net/thechance/chat/entity/Chat.kt
omer1998: great ideas, i will consider both approaches, and let you know

chat/src/main/kotlin/net/thechance/chat/entity/Chat.kt
```diff
@@ -17,4 +17,4 @@ data class Chat(
         inverseJoinColumns = [JoinColumn(name = "user_id", referencedColumnName = "id")]
     )
     val users: MutableSet<ContactUser> = mutableSetOf(),
-)
\ No newline at end of file
+    )
```
Karrar-Mohammed: extra tab that we don't need

chat/src/main/kotlin/net/thechance/chat/entity/DeletedChat.kt
```diff
@@ -0,0 +1,29 @@
...
+    @Column(name = "cleanup_status", nullable = false)
+    @Enumerated(EnumType.STRING)
+    var cleanUpStatus : CleanUpStatus,
+    @Column(name="deleted_at")
+    val deletedAt: Instant = Instant.now()
+)
+
+enum class CleanUpStatus{
+    PENDING,
+    S3_DELETED_FAILED,
```
Karrar-Mohammed: we are we specifying in the status the name of the technology that we use to save the images "S3" i prefer to call the status in general names, like regardless of the server or the technology that we are currently using because in the future we may change it, so i suggest to call it for example "MEDIA_DELETED_FAILED"

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
```diff
@@ -15,14 +18,18 @@ interface ChatRepository : JpaRepository<Chat, UUID> {
         SELECT c
         FROM Chat c
         JOIN c.users u
-        WHERE u.id in :userIds
+        WHERE u.id in :userIds 
+            AND NOT EXISTS (
+                SELECT 1 FROM DeletedChat dc WHERE dc.chatId = c.id
+                )
```
Karrar-Mohammed, line 22: instead of making it as sub query, can we just join with the deletedChat table ? i think it will be better for the performance

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
```diff
@@ -15,14 +18,18 @@ interface ChatRepository : JpaRepository<Chat, UUID> {
...
         GROUP by c
         HAVING count(u) = :#{#userIds.size}
         """
     )
     fun findByUsersIds(userIds: Set<UUID>): Chat?
 
-    @Query("SELECT c FROM Chat c JOIN c.users u WHERE u.id = :userId")
+    @Query(""" SELECT c FROM Chat c JOIN c.users u WHERE u.id = :userId 
+        AND NOT EXISTS (SELECT 1 FROM DeletedChat dc WHERE dc.chatId = c.id) """)
```
Karrar-Mohammed: same here

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
```diff
@@ -41,5 +48,19 @@ interface ChatRepository : JpaRepository<Chat, UUID> {
     fun findUnreadCountsForChats(
         @Param("chatIds") chatIds: List<UUID>
     ): List<ChatUnreadMessagesCount>
+
+    @Modifying
+    @Query(
+        nativeQuery = true,
+        value = "DELETE FROM chat.chats WHERE chat.chats.id = :chatId"
+    )
```
Karrar-Mohammed: why do we need native query for that ? i think there is function to delete by id without needing to write query by ourself

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
omer1998: you mean left join; actually, I found this as another option, for small data sets the difference is negligible 
regarding enhancing performance I found that making index on this column will be good option

chat/src/main/kotlin/net/thechance/chat/service/scheduler/ChatCleanUpScheduler.kt
```diff
@@ -0,0 +1,107 @@
...
+                    }
+
+                    CleanUpStatus.DELETED -> Unit
+                }
+            }
+        } catch (e: Exception) {
+            println("failing of chat clean up scheduler: ${e.message}")
+        }
+
+    }
```
Karrar-Mohammed, line 48: this function is big and complicated, can we split it into smaller functions ?

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
omer1998: i found article talking about differences between the performance of both; 
this the conclusion of the article 
"We conclude, first, that using the SQL NOT EXISTS or the LEFT JOIN commands are the best choice from all performance aspects. We tried also to add an index on the joining column on both tables, where the query that uses the EXCEPT command enhanced clearly and showed better performance, besides the SQL NOT EXISTS command that still the best choice overall."
https://www.sqlshack.com/t-sql-commands-performance-comparison-not-vs-not-exists-vs-left-join-vs-except/

omer1998: i will consider putting indexes (deleted chats) on this table later on

chat/src/main/kotlin/net/thechance/chat/repository/ChatRepository.kt
```diff
@@ -62,5 +63,14 @@ interface ChatRepository : JpaRepository<Chat, UUID> {
...
     fun deleteChatUsersByChatId(chatId: UUID)
-}
 
+    @Query(
+        value = """
+            SELECT c.* FROM chat.chats c 
+            JOIN chat.deleted_chats dc ON c.id = dc.chat_id
+            WHERE dc.user_id = :userId AND dc.deleted_at > :time
+        """,
+        nativeQuery = true
```
Karrar-Mohammed: i don't think need to be native query

chat/src/main/kotlin/net/thechance/chat/service/ChatService.kt
```diff
@@ -188,8 +188,22 @@ class ChatService(
...
+        val currentChat = chatRepository.findByIdOrNull(chatId) ?: throw NotFoundException("no chat found with this id $chatId")
+        currentChat.users.forEach { user ->
+            deletedChatRepository.save(DeletedChat(
+                chatId = chatId, CleanUpStatus.PENDING,
+                userId = user.id,
+                deletedAt = Instant.now()
+            ))
+
+        }
+    }
```
Karrar-Mohammed: this will create query for each one and it's not good for performance, it's better to do it in one query

Karrar-Mohammed (approved):

dev-bilal-azzam (approved):

raouf-kamel (approved):

yousef-osama11 (approved):

PR #126, opened 2025-10-29
feature/DKN-86-create-cart-and-remove-item

ElsayedMagdy122:
## what is the PR Scope ?

Implement three endpoints:

Create Cart 
Get Cart
Remove Item from Cart

## why do we need that ?

allows a user to create a new cart for a specific dukan .
allows a user to remove a specific product from their cart.

insert or update item(POST) :  /dukan/cart/{dukanId}/items
cart items (GET): /dukan/cart/{dukanId}/items
cart info : dukan/cart/{dukanId}/info
remove product from cart : /dukan/cart/dukanId/items/productId

image
image
image


dukan/src/main/kotlin/net/thechance/dukan/api/dto/cart/CartPageResponse.kt
```diff
@@ -0,0 +1,29 @@
...
+        @field:JsonProperty("page")
+        val page: Int,
+        @field:JsonProperty("size")
+        val size: Int,
+        @field:JsonProperty("total_pages")
+        val totalPages: Int,
+        @field:JsonProperty("total_items")
+        val totalItems: Long
+    )
+}
```
haideraljawasim: why did we need this custom page response , why did't you used the generic Page class from spring boot , we already use everywhere

dukan/src/main/kotlin/net/thechance/dukan/entity/Cart.kt
```diff
@@ -0,0 +1,35 @@
...
+) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || this::class != other::class) return false
+        other as Cart
+        return id == other.id
+    }
+
+    override fun hashCode(): Int = id.hashCode()
+}
```
haideraljawasim: could you please provide information why you needed to override these functions ?

dukan/src/main/kotlin/net/thechance/dukan/entity/Cart.kt
```diff
@@ -0,0 +1,35 @@
...
+    @Id
+    @GeneratedValue(strategy = GenerationType.UUID)
+    val id: UUID = UUID.randomUUID(),
+
+    @Column(name = "user_id", nullable = true)
+    val userId: UUID,
+
+    @ManyToOne
+    @JoinColumn(name = "dukan_id", nullable = false)
+    val dukan: Dukan,
```
haideraljawasim: hmmm i think we can do here join for only the dukan id we don't need the whole dukan entity? what do you think

dukan/src/main/kotlin/net/thechance/dukan/entity/CartItem.kt
```diff
@@ -0,0 +1,45 @@
...
+    val createdAt: Instant = Instant.now(),
+) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || this::class != other::class) return false
+        other as CartItem
+        return id == other.id
+    }
+
+    override fun hashCode(): Int = id.hashCode()
```
haideraljawasim, line 44: same i don't get it why to override these functions

dukan/src/main/kotlin/net/thechance/dukan/service/CartService.kt
```diff
@@ -0,0 +1,83 @@
...
+            quantity > 0 -> cart.items.add(CartItem(cart = cart, product = product, quantity = quantity))
+        }
+    }
+
+    private fun fetchCart(userId: UUID, dukanId: UUID): Cart =
+        cartRepository.findByUserIdAndDukanId(userId, dukanId)
+            ?: throw CartNotFoundException()
+
+    fun fetchCartWithPagedItems(userId: UUID, dukanId: UUID, pageable: Pageable): CartPageResponse {
+        val cart = fetchCart(userId, dukanId)
```
haideraljawasim: lets keep the same naming convention, use `get` instead of `fetch`

dukan/src/main/kotlin/net/thechance/dukan/service/CartService.kt
```diff
@@ -0,0 +1,83 @@
...
+            existingItem != null -> existingItem.quantity = quantity
+            quantity > 0 -> cart.items.add(CartItem(cart = cart, product = product, quantity = quantity))
+        }
+    }
+
+    private fun fetchCart(userId: UUID, dukanId: UUID): Cart =
+        cartRepository.findByUserIdAndDukanId(userId, dukanId)
+            ?: throw CartNotFoundException()
+
+    fun fetchCartWithPagedItems(userId: UUID, dukanId: UUID, pageable: Pageable): CartPageResponse {
```
haideraljawasim: here what i mean you can use `Page<Cart>`
instead of this custom page response

dukan/src/main/kotlin/net/thechance/dukan/service/CartService.kt
```diff
@@ -0,0 +1,83 @@
...
+        val cart = fetchCart(userId, dukanId)
+        val itemsPage = cartItemRepository.findAllByCartId(cart.id, pageable)
+        val itemsResponse = itemsPage.content.map { it.toResponse() }
+        val totalPrice = cart.items.sumOf { it.product.price * it.quantity }
+
```
haideraljawasim: also the mapping should be inside the controller for the response

dukan/src/main/kotlin/net/thechance/dukan/entity/SoldProduct.kt
```diff
@@ -0,0 +1,43 @@
...
+    @GeneratedValue(strategy = GenerationType.UUID)
+    val id: UUID = UUID.randomUUID(),
+
+    @ManyToOne
+    @JoinColumn(name = "product_id", nullable = false)
+    val product: DukanProduct,
+
+    @ManyToOne
+    @JoinColumn(name = "dukan_id", nullable = false)
+    val dukan: Dukan,
```
haideraljawasim: hmmm , same here why don't we go only with id of dukan and id of the product no need to whole dukan or product

dukan/src/main/kotlin/net/thechance/dukan/service/CartService.kt
```diff
@@ -0,0 +1,83 @@
...
+        )
+    }
+
+    @Transactional
+    fun removeItem(userId: UUID, dukanId: UUID, productId: UUID) {
+        val cart = fetchCart(userId, dukanId)
+        val item = cart.items.find { it.product.id == productId } ?: throw ProductNotInCartException()
+        cart.items.remove(item)
+        cartRepository.save(cart)
+    }
```
haideraljawasim, line 87: you can do this with query to find the exact item and delete it no need to fetch the whole cart and loop through it

dukan/src/main/kotlin/net/thechance/dukan/entity/CartItem.kt
```diff
@@ -0,0 +1,45 @@
...
+data class CartItem(
+    @Id
+    @GeneratedValue(strategy = GenerationType.UUID)
+    val id: UUID = UUID.randomUUID(),
+
+    @ManyToOne
+    @JoinColumn(name = "cart_id", nullable = false)
+    @JsonIgnore
+    val cart: Cart,
+
```
haideraljawasim, line 26: I feel like the product itself should have a quantity field, and we don’t really need a separate cart item anymore. every time we fetch a product, we can include the quantity that the logged in user has in their cart. I’m thinking about a scenario where, on the product screen, we want to display how many items of that product the user already has in their cart , which what we have in our case

haideraljawasim (changes requested):
We can discuss this more in our daily meeting

dukan/src/main/kotlin/net/thechance/dukan/service/CartService.kt
ElsayedMagdy122: That's it, I'll return the paging cart, and I want to return the paging elements.

dukan/src/main/kotlin/net/thechance/dukan/api/dto/cart/CartPageResponse.kt
ElsayedMagdy122: That's it, I'll return the paging cart, and I want to return the paging elements.

dukan/src/main/kotlin/net/thechance/dukan/entity/Cart.kt
ElsayedMagdy122: you are right

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
```diff
@@ -0,0 +1,42 @@
...
+import java.util.*
+
+@RestController
+@RequestMapping("$DUKAN_PATH/cart")
+class CartController(
+    private val cartService: CartService
+) {
+    @PostMapping("/items")
+    fun addOrUpdateItem(
+        @AuthenticationPrincipal userId: UUID, @RequestBody request: AddToCartRequest
```
TarekIdrees: it seems you forget to add  _**@valid**_ annotation here before _**@RequestBoday**_ annotationIo

dukan/src/main/kotlin/net/thechance/dukan/service/CartService.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    private fun findOrCreateCart(userId: UUID, dukanId: UUID): Cart {
```
TarekIdrees: The function name seems to do more than one thing :D
It would be better if you isolate each functionality so you can use find or create for diiferent use cases. f

dukan/src/main/kotlin/net/thechance/dukan/service/CartService.kt
```diff
@@ -0,0 +1,83 @@
...
+
+@Service
+class CartService(
+    private val cartRepository: CartRepository,
+    private val productRepository: DukanProductRepository,
+    private val dukanRepository: DukanRepository,
+    private val cartItemRepository: CartItemRepository,
+) {
+    @Transactional
+    fun addOrUpdateItem(userId: UUID, dukanId: UUID, productId: UUID, quantity: Int): Cart {
```
TarekIdrees: It would be nice if you could combine all these parameters using one data class. Check other services for your reference.

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
```diff
@@ -0,0 +1,42 @@
...
+        return ResponseEntity.ok().build()
+    }
+
+    @GetMapping("/{dukanId}")
+    fun getCart(
+        @AuthenticationPrincipal userId: UUID,
+        @PathVariable dukanId: UUID,
+        @PageableDefault(size = 10, page = 0, sort = ["createdAt"], direction = Sort.Direction.DESC) pageable: Pageable
+    ): ResponseEntity<CartPageResponse> {
+        val cartResponse = cartService.fetchCartWithPagedItems(userId, dukanId, pageable)
```
TarekIdrees: How about using _**get()**_ instead of **_fetch()_** for more consistency with the rest of the code?

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
```diff
@@ -0,0 +1,42 @@
...
+        return ResponseEntity.ok(cartResponse)
+    }
+
+    @DeleteMapping("/{dukanId}/items/{productId}")
+    fun removeItem(@AuthenticationPrincipal userId: UUID, @PathVariable dukanId: UUID, @PathVariable productId: UUID) {
+        ResponseEntity.ok(cartService.removeItem(userId, dukanId, productId))
```
TarekIdrees: Better if we sperate it into two lines also **_ResponseEntity.noContent().build()_** ideal for DELETE .

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @GetMapping("/{dukanId}/info")
+    fun getCartInfo(
+        @AuthenticationPrincipal userId: UUID,
+        @PathVariable dukanId: UUID
+    ): ResponseEntity<CartResponse> {
+        val cart = cartService.getCart(userId, dukanId)
+        val totalPrice = cart.items.sumOf { it.product.price * it.quantity }
+        return ResponseEntity.ok(cart.toResponse(totalPrice))
```
TarekIdrees: Isolate the step of mapping one line before

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
```diff
@@ -0,0 +1,65 @@
...
+
+    @GetMapping("/{dukanId}/items")
+    fun getCartItems(
+        @AuthenticationPrincipal userId: UUID,
+        @PathVariable dukanId: UUID,
+        @PageableDefault(size = 10, sort = ["createdAt"], direction = Sort.Direction.DESC)
+        pageable: Pageable
+    ): ResponseEntity<Page<CartItemResponse>> {
+        val items = cartService.getCartItems(userId, dukanId, pageable)
+        return ResponseEntity.ok(items.map { it.toResponse() })
```
TarekIdrees: Same as well here

dukan/src/main/kotlin/net/thechance/dukan/service/CartService.kt
```diff
@@ -0,0 +1,74 @@
...
+            ?: throw CartNotFoundException()
+    }
+
+    @Transactional(readOnly = true)
+    fun getCartItems(userId: UUID, dukanId: UUID, pageable: Pageable): Page<CartItem> {
+        val cart = getCart(userId, dukanId)
+        return cartItemRepository.findAllByCartId(cart.id, pageable)
+    }
+
+    private fun getOrCreateCart(userId: UUID, dukanId: UUID): Cart {
```
TarekIdrees: I don't recommend this kind of functions which has two or more responsibilities

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
```diff
@@ -0,0 +1,65 @@
...
+class CartController(
+    private val cartService: CartService
+) {
+
+    @PostMapping("/items")
+    fun addOrUpdateItem(
+        @AuthenticationPrincipal userId: UUID,
+        @Valid @RequestBody request: AddToCartRequest
+    ): ResponseEntity<Unit> {
+        cartService.addOrUpdateItem(params = AddOrUpdateCartItemParams(userId, request.dukanId, request.productId, request.quantity))
```
YoussefmSaber: enhancement for readability:
make every param in a line so this function call is not too long
smth like this
```
AddOrUpdateCartItemParams(
userId,
 request.dukanId, 
 request.productId,
  request.quantity
  )

File: dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanProductController.kt [hunk 20]
```diff
@@ -49,18 +49,20 @@ class DukanProductController(
...
+        val products = dukanProductService.getProductsByShelf(userId,shelfId, pageable)
+        return ResponseEntity.ok(products)
     }
 
     @GetMapping("/{productId}")
-    fun getProductById(@PathVariable("productId") productId: UUID): ResponseEntity<DukanProductResponse> {
-        val productResponse = dukanProductService.getProductById(productId).toProductResponse()
+    fun getProductById(
+        @AuthenticationPrincipal userId: UUID,
+        @PathVariable("productId") productId: UUID): ResponseEntity<DukanProductResponse> {
```
YoussefmSaber: make the return type in a new line, so the code be unified

dukan/src/main/kotlin/net/thechance/dukan/entity/SoldProduct.kt
YoussefmSaber: Just store a reference for dukan_id, and product_id, you don't need to store a whole entity just for reducing queries.

dukan/src/main/kotlin/net/thechance/dukan/entity/CartItem.kt
```diff
@@ -0,0 +1,45 @@
...
+    @ManyToOne
+    @JoinColumn(name = "product_id", nullable = false)
+    val product: DukanProduct,
```
YoussefmSaber, line 29: same here, just make reference using the id, and use the product id as a foreign key

dukan/src/main/kotlin/net/thechance/dukan/repository/CartRepository.kt
```diff
@@ -0,0 +1,18 @@
...
+import org.springframework.data.jpa.repository.Query
+import java.util.*
+
+interface CartRepository : JpaRepository<Cart, UUID> {
+    @Query(
+        """
+        select c from Cart c
+        join fetch c.items i
+        join fetch i.product p
+        where c.userId = :userId and c.dukanId = :dukanId
```
YoussefmSaber: don't just use letters for naming, use normal naming, like cart, items, product.

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanProductRepository.kt
```diff
@@ -5,12 +5,33 @@ import org.springframework.data.domain.Page
...
 interface DukanProductRepository : JpaRepository<DukanProduct, UUID> {
     fun existsByDukanIdAndNameIgnoreCase(dukanId: UUID, name: String): Boolean
-    fun findAllByShelfId(shelfId: UUID, pageable: Pageable): Page<DukanProduct>
     fun existsByShelfId(shelfId: UUID): Boolean
     fun findByIdAndDukanOwnerId(id: UUID, ownerId: UUID): Optional<DukanProduct>
+    @Query("""
+        select p from DukanProduct p
+        join fetch p.shelf s
+        join fetch s.dukan d
+        where s.id = :shelfId
```
YoussefmSaber: same thing here

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
```diff
@@ -0,0 +1,65 @@
...
+        return ResponseEntity.ok().build()
+    }
+
+    @GetMapping("/{dukanId}/info")
+    fun getCartInfo(
+        @AuthenticationPrincipal userId: UUID,
+        @PathVariable dukanId: UUID
+    ): ResponseEntity<CartResponse> {
+        val cart = cartService.getCart(userId, dukanId)
+        val totalPrice = cart.items.sumOf { it.product.price * it.quantity }
```
haideraljawasim: this sum logic should be inside a service or gettin using query, not here in controller

dukan/src/main/kotlin/net/thechance/dukan/api/dto/cart/CartResponse.kt
```diff
@@ -0,0 +1,11 @@
+package net.thechance.dukan.api.dto.cart
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import java.util.*
+
+data class CartResponse(
+    @field:JsonProperty("id")
+    val id: UUID,
+    @field:JsonProperty("total_price")
+    val totalPrice: Double,
```
haideraljawasim, line 8: hmmmm, I think for the naming of the response fields we used camelCase  everywhere , i prefer keeping the same naming convention

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -71,16 +75,28 @@ class DukanProductService(
...
+        val cart = cartRepository.findByUserIdAndDukanIdWithItemsAndProducts(userId, shelf.dukan.id)
+        val quantityMap = cart?.items?.associate { it.product.id to it.quantity } ?: emptyMap()
 
-    fun getProductById(productId: UUID): DukanProduct {
-        return dukanProductRepository.findById(productId).orElseThrow {
-            ProductNotFoundException()
+        return productsPage.map { product ->
+            val quantity = quantityMap[product.id] ?: 0
+            product.toProductResponse(quantity)
         }
```
haideraljawasim, line 82: Please return this as it used to be first we return entity from service and then map it to response inside the controller, second i don't get the changes you did here , why would we need to check the shelf exist inside the response or not also what is the zero here you send inside the to response mapper , this code need more explanation and reasoning for why we went with this approach

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanProductRepository.kt
```diff
@@ -5,12 +5,33 @@ import org.springframework.data.domain.Page
...
+        pageable: Pageable
+    ): Page<DukanProduct>
+
+    @Query("""
+        select p from DukanProduct p
+        join fetch p.shelf s
+        join fetch s.dukan d
+        where p.id = :productId
+    """)
+    fun findByIdWithDukan(
```
haideraljawasim: i think you should go with another approach here  rather than `join fetch` as i read the` join fetch` here will latterly fetch the entity  and its relations in one single query like using `FetchType.Eager `  which might be harmful for the performance cause it will need to store the relations graph  in RAM , also im not sure as read that spring boot still not support pagination with join fetch , search about  DTO projections it might be useful in our case

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -71,16 +75,28 @@ class DukanProductService(
...
     }
+    @Transactional
+    fun getProductById(userId: UUID, productId: UUID): DukanProductResponse {
+        val product = dukanProductRepository.findByIdWithDukan(productId)
+        val cart = cartRepository.findByUserIdAndDukanIdWithItemsAndProducts(userId, product.shelf.dukan.id)
+        val quantity = cart?.items?.find { it.product.id == product.id }?.quantity ?: 0
+
+        return product.toProductResponse(quantity)
+    }
```
haideraljawasim, line 93: we already talked that a cart can contain hundreds or thousands of items since there’s no limit on how many a user can add. fetching the entire cart with all its items and then mapping through them would require storing everything in the server’s RAM, which isn’t the most best  approach. please look for another way to retrieve the products along with their quantities using a query

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -71,16 +75,28 @@ class DukanProductService(
...
-    fun getProductById(productId: UUID): DukanProduct {
-        return dukanProductRepository.findById(productId).orElseThrow {
-            ProductNotFoundException()
+        return productsPage.map { product ->
+            val quantity = quantityMap[product.id] ?: 0
+            product.toProductResponse(quantity)
         }
     }
+    @Transactional
+    fun getProductById(userId: UUID, productId: UUID): DukanProductResponse {
```
haideraljawasim: another thing i don't get it why would you change the service to return a response this controller job the service should return an entity or output class !

haideraljawasim (changes requested):
Great job , but still need some enhancements

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
```diff
@@ -0,0 +1,65 @@
...
+    @PostMapping("/items")
+    fun addOrUpdateItem(
+        @AuthenticationPrincipal userId: UUID,
+        @Valid @RequestBody request: AddToCartRequest
+    ): ResponseEntity<Unit> {
+        cartService.addOrUpdateItem(params = AddOrUpdateCartItemParams(userId, request.dukanId, request.productId, request.quantity))
+        return ResponseEntity.ok().build()
+    }
+
+    @GetMapping("/{dukanId}/info")
```
mostafa-n3ma, line 48: The endpoint get dukan/cart/{dukanid} is enough I think we don't need the /info

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
mostafa-n3ma: Agree with haider we should do the sum logic from the service layer

mostafa-n3ma (approved):

YoussefmSaber (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanProductController.kt
```diff
@@ -49,18 +49,21 @@ class DukanProductController(
...
     fun getProductsByShelf(
+        @AuthenticationPrincipal userId: UUID,
         @RequestParam shelfId: UUID,
         @PageableDefault(size = 10, page = 0, sort = ["createdAt"], direction = Sort.Direction.DESC)
         pageable: Pageable
     ): ResponseEntity<Page<DukanProductResponse>> {
-        val products = dukanProductService.getProductsByShelf(shelfId, pageable)
-        val productsResponse = products.map { it.toProductResponse() }
-        return ResponseEntity.ok(productsResponse)
+        val products = dukanProductService.getProductsByShelf(userId, shelfId, pageable)
```
TarekIdrees, line 53: The mapping step should happen here in the controller, so I would recommend reverting these changes.

dukan/src/main/kotlin/net/thechance/dukan/api/controller/DukanProductController.kt
```diff
@@ -49,18 +49,21 @@ class DukanProductController(
...
 
     @GetMapping("/{productId}")
-    fun getProductById(@PathVariable("productId") productId: UUID): ResponseEntity<DukanProductResponse> {
-        val productResponse = dukanProductService.getProductById(productId).toProductResponse()
+    fun getProductById(
+        @AuthenticationPrincipal userId: UUID,
+        @PathVariable("productId") productId: UUID
+    ): ResponseEntity<DukanProductResponse> {
+        val productResponse = dukanProductService.getProductById(userId, productId)
         return ResponseEntity.ok(productResponse)
```
TarekIdrees, line 65: Here as well, mapping step

AbdelrhmanAyman20470 (approved):

TarekIdrees (approved):
Well done!

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
```diff
@@ -0,0 +1,73 @@
...
+        cartService.addOrUpdateItem(
+            params = AddOrUpdateCartItemParams(
+                userId,
+                request.dukanId,
+                request.productId,
+                request.quantity
+            )
+        )
+        return ResponseEntity.ok().build()
+    }
```
devfalah, line 46: let's create a new endpoint for update functionality

dukan/src/main/kotlin/net/thechance/dukan/api/dto/cart/CartItemResponse.kt
```diff
@@ -0,0 +1,19 @@
...
+    @field:JsonProperty("product_name")
+    val productName: String,
+    @field:JsonProperty("description")
+    val description: String,
+    @field:JsonProperty("quantity")
+    val quantity: Int,
+    @field:JsonProperty("price")
+    val price: Double,
+    @field:JsonProperty("image_url")
+    val imageUrl: String?
```
devfalah, line 12: I see an inconsistency in your PR here you use snake_case, while in other responses you use camelCase

devfalah (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/controller/CartController.kt
ElsayedMagdy122: done

dukan/src/main/kotlin/net/thechance/dukan/api/dto/cart/CartItemResponse.kt
ElsayedMagdy122: done

YoussefmSaber (approved):

haideraljawasim (approved):

PR #130, opened 2025-10-29
fix: fix delete when uploading and add details endpoint for my trends screen

mohamedd-hassan:
## what is the PR Scope ?
fix behavior in deleting a trend in the upload screen before it has a thunbnail would fail so i added a null check before it attempts to delete a trend without a thumbnail (which is valid). also added an extra optional parameter for the get my trends endpoint so that the mobile can use it

## why do we need that ?
to fix incorrect behavior when calling these endpoints

`trends/user/trendId` same response just narrower results

### DIFF
--- a/trends/src/main/kotlin/net/thechance/trends/api/controller/TrendUserController.kt
+++ b/trends/src/main/kotlin/net/thechance/trends/api/controller/TrendUserController.kt
@@ -8,6 +8,7 @@ import org.springframework.data.domain.Pageable
 import org.springframework.http.ResponseEntity
 import org.springframework.security.core.annotation.AuthenticationPrincipal
 import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.RequestMapping
 import org.springframework.web.bind.annotation.RestController
 import java.util.*
@@ -18,13 +19,14 @@ class TrendUserController(
     private val trendsService: TrendsService
 ) {
 
-    @GetMapping
+    @GetMapping("/{trendId}", "")
     fun getAllTrendsByUserId(
         pageable: Pageable,
+        @PathVariable(required = false) trendId: UUID? = null,
         @AuthenticationPrincipal currentUserId: UUID
     ): ResponseEntity<PagingResponse<TrendResponse>> {
 
-        val trends = trendsService.getAllTrendsByUserId(pageable, currentUserId).content.map { it.toResponse() }
+        val trends = trendsService.getAllTrendsByUserId(pageable, currentUserId, trendId).content.map { it.toResponse() }
 
         val result = PagingResponse(
             pageNumber = pageable.pageNumber,
--- a/trends/src/main/kotlin/net/thechance/trends/repository/TrendsRepository.kt
+++ b/trends/src/main/kotlin/net/thechance/trends/repository/TrendsRepository.kt
@@ -22,15 +22,21 @@ interface TrendsRepository : JpaRepository<Trend, UUID> {
         LEFT JOIN TrendLike tl ON tl.trendId = t.id AND tl.userId = :ownerId
         WHERE t.ownerId = :ownerId 
         AND t.isPublished = :isPublished
+        AND (:trendId IS NULL OR t.createdAt <= (
+            SELECT t2.createdAt FROM Trend t2 WHERE t2.id = :trendId
+        ))
         """,
         countQuery = """
         SELECT COUNT(t.id)
         FROM Trend t
         WHERE t.ownerId = :ownerId 
         AND t.isPublished = :isPublished
+        AND (:trendId IS NULL OR t.createdAt <= (
+            SELECT t2.createdAt FROM Trend t2 WHERE t2.id = :trendId
+        ))
         """
     )
-    fun findByOwnerIdAndIsPublished(ownerId: UUID, isPublished: Boolean, pageable: Pageable): Page<TrendWithLikeStatus>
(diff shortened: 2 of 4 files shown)


Karrar-Mohammed:
it's better to mention what is the correct behavior in the PR description and how you solve it

trends/src/main/kotlin/net/thechance/trends/api/controller/TrendUserController.kt
```diff
@@ -18,13 +19,14 @@ class TrendUserController(
     private val trendsService: TrendsService
 ) {
 
-    @GetMapping
+    @GetMapping("/{trendId}", "")
     fun getAllTrendsByUserId(
         pageable: Pageable,
+        @PathVariable(required = false) trendId: UUID? = null,
```
Karrar-Mohammed, line 25: why are we passing the trendId here ? and why it's nullable ?

trends/src/main/kotlin/net/thechance/trends/api/controller/TrendUserController.kt
```diff
@@ -18,13 +19,14 @@ class TrendUserController(
     private val trendsService: TrendsService
 ) {
 
-    @GetMapping
+    @GetMapping("/{trendId}", "")
```
Karrar-Mohammed, line 22: what is this second parameter ? why do we need to pass it ? and why are we passing it as empty string ?

trends/src/main/kotlin/net/thechance/trends/repository/TrendsRepository.kt
```diff
@@ -22,15 +22,21 @@ interface TrendsRepository : JpaRepository<Trend, UUID> {
...
         FROM Trend t
         WHERE t.ownerId = :ownerId 
         AND t.isPublished = :isPublished
+        AND (:trendId IS NULL OR t.createdAt <= (
+            SELECT t2.createdAt FROM Trend t2 WHERE t2.id = :trendId
+        ))
         """
     )
-    fun findByOwnerIdAndIsPublished(ownerId: UUID, isPublished: Boolean, pageable: Pageable): Page<TrendWithLikeStatus>
+    fun findByOwnerIdAndIsPublished(ownerId: UUID, isPublished: Boolean, trendId: UUID?, pageable: Pageable): Page<TrendWithLikeStatus>
```
Karrar-Mohammed, line 39: this function is scary and i don't think we need it to be this complicated i'm sure there is simpler way to achieve the same result

Karrar-Mohammed (commented):
the PR title and description is about fixing wrong behavior in deleting a trend, but i don't see any changes related to delete,

all i see is just unreasonable code and not related to the PR title

trends/src/main/kotlin/net/thechance/trends/api/controller/TrendUserController.kt
mohamedd-hassan: Because its an optional parameter, if they dont provide it the endpoint will return all trends, if the provide an id it will return all trends starting from that Id

trends/src/main/kotlin/net/thechance/trends/api/controller/TrendUserController.kt
mohamedd-hassan: previously it was an empty get mapping, but to make the path parameter (trendId) optional we had to give another endpoint in case the trendId is null and since we want to have an empty get mapping we have to make that second parameter empty

trends/src/main/kotlin/net/thechance/trends/repository/TrendsRepository.kt
mohamedd-hassan: this function has to be a manual query because it returns something based on current userId (isLiked) and based on the nullable paramter we either get larger than or equal to that id

mohamedd-hassan:
> the PR title and description is about fixing wrong behavior in deleting a trend, but i don't see any changes related to delete,
> 
> all i see is just unreasonable code and not related to the PR title

its about fixing deleting trend (null check) AND adding details to that get my trends endpoint

trends/src/main/kotlin/net/thechance/trends/api/controller/TrendUserController.kt
Karrar-Mohammed: and why we need that ?

trends/src/main/kotlin/net/thechance/trends/repository/TrendsRepository.kt
Karrar-Mohammed: what do you mean by larger than or equal to id ? can you explain it more and why we need that ?

Karrar-Mohammed:
ok after you updated the description now i can understand the delete issue and how you fixed it, but i'm still don't see passing the trend id as nullable to the end point that need to return all the trends, i see it can be another end point, maybe i need to understand the business requirement more

mohamedd-hassan:
> ok after you updated the description now i can understand the delete issue and how you fixed it, but i'm still don't see passing the trend id as nullable to the end point that need to return all the trends, i see it can be another end point, maybe i need to understand the business requirement more

you know how some endpoints have an optional date parameter where if you dont pass it it returns all results but if you do it returns results starting from that date? here its the same exact case but with TrendId, if we pass it we return everything starting from that Id and if we dont we return everything

trends/src/main/kotlin/net/thechance/trends/repository/TrendsRepository.kt
mohamedd-hassan: its basically the same as fetching everything after the current date but instead of date its an Id as all the trends in the user profile are sequential and we need to fetch from a certain date sometimes

Karrar-Mohammed (approved):

ahmeddsamii (approved):

iZyadNasser (approved):

PR #132, opened 2025-10-30
Feature/fth 104 design and create mosques database table

Hasa9nAli:
what is the PR Scope ?
create table , request, response structure ، exceptions , and mapper 

why do we need that ?
allow us to create new end point in the next feature wanted like get mosques , create new mosque , search about mosque 

end points with the request and response body:
without end point controller


app/src/main/kotlin/net/thechance/config/storage/StorageConfig.kt
```diff
@@ -27,6 +27,11 @@ class StorageConfig(
     fun trendsS3Client(trendCreds: StaticCredentialsProvider): S3Client =
         buildClient(props.trends.endpoint, trendCreds)
 
+
+    @Bean
+    fun faithS3Client(): S3Client =
+        buildClient(props.trends.endpoint, trendCreds())
```
Karrar-Mohammed: why are we using trends end point and trendCreds here ?

app/src/main/resources/application.properties
```diff
@@ -39,11 +39,15 @@ storage.wallet.bucket=${STORAGE_WALLET_BUCKET}
 storage.wallet.endpoint=${STORAGE_WALLET_ENDPOINT}
 storage.wallet.cdn-endpoint=${STORAGE_WALLET_CDN_ENDPOINT}
 
+storage.faith.key=${STORAGE_FAITH_KEY}
+storage.faith.secret=${STORAGE_FAITH_SECRET}
+storage.faith.bucket=${STORAGE_FAITH_BUCKET}
+storage.faith.endpoint=${STORAGE_FAITH_ENDPOINT}
+storage.faith.cdn-endpoint=${STORAGE_FAITH_CDN_ENDPOINT}
```
Karrar-Mohammed: as i remember falah mentioned that we will have storage for dukan, trends, and wallet and the other modules can use the mena storage, so why are we adding configuration for new storage ?

app/src/main/kotlin/net/thechance/config/storage/StorageConfig.kt
Mostafa-khaled21: I’ve updated it according to your previous comment — we’re now using Mena storage instead of creating a new one.

I removed the extra storage configuration and used the Mena storage in the FaithImageStorageService file under the Faith module.

app/src/main/resources/application.properties
Mostafa-khaled21: I’ve updated the code, please take another look.
@Karrar-Mohammed

devfalah (approved):

faith/src/main/kotlin/net/thechance/faith/api/dto/error/ImageExcetion.kt
```diff
@@ -0,0 +1,20 @@
+package net.thechance.faith.api.controller.exception
+
+import org.springframework.http.HttpStatus
+
+
+const val INVALID_IMAGE_FORMAT = 1201
+const val IMAGE_UPLOAD_FAILED = 1202
+
+
+class InvalidImageFormatException : FaithException(
```
marwanqashwa: i think we don't use code and status  in exeption we just throws it and handle statues and code in exeption handler

Mostafa-khaled21: I’ve updated the code, please take another look.
@marwanqashwa

marwanqashwa (approved):

mohannadahmed00 (approved):

PR #133, opened 2025-10-30
Feature/IDN-130 Add change password Endpoint

eslam-99:
## what is the PR Scope ?

The scope is adding change password Endpoint

## why do we need that ?

- `/identity/profile/change-password` to allow user to change their password

- Endpoint: `/identity/profile/change-password`

Curl:
```
curl --location 'https://mena-dev.the-chance.net/identity/profile/change-password' \
--header 'Content-Type: application/json' \
--header 'Authorization: Bearer ' \
--data '{
    "currentPassword":"Abcd1234",
    "newPassword":"12345678",
    "confirmPassword":"12345678"
}'
```

RequestBody: 
```json
{
    "currentPassword":"Abcd1234",
    "newPassword":"12345678",
    "confirmPassword":"12345678"
}
```

Expected returned body:
```json
{
    "message": "Password changed successfully"
}
```

Expected exceptions

- 401 (Unauthorized): If current password not matches the password in DB
- 400 (Bad Request): If request has invalid data


identity/src/main/kotlin/net/thechance/identity/api/controller/ProfileControllerAdvice.kt
```diff
@@ -2,21 +2,40 @@ package net.thechance.identity.api.controller
...
+    @ExceptionHandler(MethodArgumentNotValidException::class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    fun handleValidationExceptions(exception: MethodArgumentNotValidException): ResponseEntity<ErrorResponse?> {
+        val firstException = exception.getFirstException()
+        logger.error("Validation failed: $firstException", exception)
+        return ResponseEntity
+            .status(HttpStatus.BAD_REQUEST)
+            .body(ErrorResponse(firstException))
+    }
+
```
poula22, line 36: Here we are sending to user what is wrong request and this can be away of telling how to send valid request to our system from outside application

identity/src/main/kotlin/net/thechance/identity/service/ChangePasswordService.kt
```diff
@@ -0,0 +1,49 @@
...
+) {
+    fun changePassword(userId: UUID, currentPassword: String, newPassword: String, confirmPassword: String) {
+        runCatching {
+            validateRequest(newPassword, confirmPassword)
+            findAndVerifyUser(userId, currentPassword)
+        }.map { verifiedUser ->
+            updateUserWithNewPassword(verifiedUser, newPassword)
+        }.onSuccess { updatedUser ->
+            userRepository.save(updatedUser)
+        }.getOrThrow()
```
poula22: runCatching then getOrThrow? run catching has no value here

identity/src/test/kotlin/net/thechance/identity/service/ChangePasswordServiceTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+        val USER = createUser(
+            USER_ID,
+            password = CURRENT_ENCODED_PASSWORD
+        )
+
+        val UPDATED_USER = createUser(
+            USER_ID,
+            password = NEW_ENCODED_PASSWORD
+        )
```
poula22, line 99: I have a suggestion here to use Users in test utilities. This will give us more data for testing in features we can use then.

poula22 (changes requested):

identity/src/main/kotlin/net/thechance/identity/api/controller/ProfileControllerAdvice.kt
```diff
@@ -2,21 +2,40 @@ package net.thechance.identity.api.controller
...
     private val logger: Logger = LoggerFactory.getLogger(ProfileController::class.java)
 
+    @ExceptionHandler(MethodArgumentNotValidException::class)
+    @ResponseStatus(HttpStatus.BAD_REQUEST)
+    fun handleValidationExceptions(exception: MethodArgumentNotValidException): ResponseEntity<ErrorResponse?> {
+        val firstException = exception.getFirstException()
+        logger.error("Validation failed: $firstException", exception)
+        return ResponseEntity
+            .status(HttpStatus.BAD_REQUEST)
+            .body(ErrorResponse(firstException))
```
ThoriaHamdy99: I think, we can send here error message like Data not valid or something like that instead of sending one error message and drop the others. or what do you think

identity/src/main/kotlin/net/thechance/identity/api/dto/ChangePassword.kt
```diff
@@ -0,0 +1,15 @@
+package net.thechance.identity.api.dto
+
+import org.hibernate.validator.constraints.Length
+
+data class ChangePasswordRequest(
+    val currentPassword: String,
```
ThoriaHamdy99, line 7: why there is no validation on the current password

identity/src/main/kotlin/net/thechance/identity/service/ChangePasswordService.kt
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    private fun validateRequest(newPassword: String, confirmPassword: String) {
+        (newPassword == confirmPassword).takeIf { it } ?: throw PasswordMismatchException()
```
ThoriaHamdy99: Are we need another check about the current password or what?

ThoriaHamdy99: Okay you made it below

devfalah (approved):

identity/src/main/kotlin/net/thechance/identity/api/controller/ProfileControllerAdvice.kt
eslam-99: Done

identity/src/main/kotlin/net/thechance/identity/service/ChangePasswordService.kt
eslam-99: Done

identity/src/test/kotlin/net/thechance/identity/service/ChangePasswordServiceTest.kt
eslam-99: No need as these data are only for these tests

identity/src/main/kotlin/net/thechance/identity/api/dto/ChangePassword.kt
eslam-99: Done

ThoriaHamdy99 (approved):

poula22 (approved):

PR #136, opened 2025-10-31
Fix/wlt 106 statement generation for large transaction amount

MuhammedEdrees:
## what is the PR Scope ?
- Fixed a bug in opening balance calculation.
- Fixed a bug where statement generation fails for statements with more than 100 transactions.

### DIFF
--- a/wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
+++ b/wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
@@ -38,13 +38,15 @@ class StatementService(
             types = types,
             startDateTime = startDateTime,
             endDateTime = endDateTime,
-            openingBalance = getOpeningBalance(userId, startDate),
-            closingBalance = getClosingBalance(userId, endDate)
+            openingBalance = getOpeningBalance(userId, startDateTime),
+            closingBalance = getClosingBalance(userId, endDateTime)
         )
     }
 
     private fun getStartDateTime(startDate: LocalDate?, userId: UUID): LocalDateTime {
-        return startDate?.atStartOfDay() ?: transactionService.getUserFirstTransactionDate(userId).orNow()
+        val dateTime = startDate?.atStartOfDay()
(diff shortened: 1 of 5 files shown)


wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
```diff
@@ -38,13 +38,15 @@ class StatementService(
...
+            openingBalance = getOpeningBalance(userId, startDateTime),
+            closingBalance = getClosingBalance(userId, endDateTime)
         )
     }
 
     private fun getStartDateTime(startDate: LocalDate?, userId: UUID): LocalDateTime {
-        return startDate?.atStartOfDay() ?: transactionService.getUserFirstTransactionDate(userId).orNow()
+        val dateTime = startDate?.atStartOfDay()
+            ?: transactionService.getUserFirstTransactionDate(userId).orNow()
+        return dateTime.toLocalDate().atStartOfDay()
```
Joseph-Sameh-0: why you convert it to date and take the start of the date as date time ?

Joseph-Sameh-0 (approved):

Malak187 (approved):

wallet/src/main/kotlin/net/thechance/wallet/service/StatementService.kt
MuhammedEdrees: To include all the transaction in this date from the beginning of the day as this is the start date and for the end date it's at the end of the day.

Joseph-Sameh-0:
```suggestion
        return startDate?.atStartOfDay()
            ?: transactionService.getUserFirstTransactionDate(userId)
                .orNow().toLocalDate().atStartOfDay()
```
I suggest to make it like this t avoid multiple mapping if the startDate not null

MuhammedEdrees: Ok makes sense

Karrar-Mohammed (approved):

PR #141, opened 2025-10-31
Feature/wlt 89 get users endpoint

MuhammedEdrees:
## what is the PR Scope ?
- Add search users endpoint

## why do we need that ?
- To have control and monitoring over users in admin panel.

### 1- Get Users:
**Request**: 
```HTTP
GET /identity/admin/users?query=john&page=0&size=10&sort=firstName,asc
```
**Response**:
```JSON
{
  "totalElements": 1,
  "page": 0,
  "pageSize": 10,
  "totalPages": 1,
  "items": [
    {
      "id": "b5b1f9b4-2d44-4cb8-8d4b-3a8b1b6c42f7",
      "firstName": "John",
      "lastName": "Doe",
      "phoneNumber": "+201012345678",
      "lastLoginAt": "2025-10-30T14:22:15",
      "lastVisitAt": "2025-10-30T14:25:41",
      "status": "ACTIVE"
    }
  ]
}
```

### 2- Update User Status:
**Request**: 
```HTTP
PATCH /identity/admin/users/{userId}/status

{
  "status": "BLOCKED"
}
```

**Response**:  `200 OK`


identity/src/main/kotlin/net/thechance/identity/api/dto/UserResponse.kt
```diff
@@ -0,0 +1,15 @@
+package net.thechance.identity.api.dto
+
+import net.thechance.identity.entity.User
+import java.time.LocalDateTime
+import java.util.*
+
+data class UserResponse(
+    val id: UUID,
```
Joseph-Sameh-0: what the UserResponse means ?
can you try to rename it better ?

identity/src/main/kotlin/net/thechance/identity/entity/User.kt
```diff
@@ -41,9 +39,24 @@ data class User(
     @field:Range(min = Gender.MALE, max = Gender.FEMALE)
     @Column(name = "gender",  nullable = false)
     val gender: Int,
+
+    @Column(name = "last_login_at", nullable = false)
+    val lastLoginAt: LocalDateTime = LocalDateTime.now(),
+
```
Joseph-Sameh-0, line 45: you can't take the last login from login log table ?

Malak187 (approved):
💯

identity/src/main/kotlin/net/thechance/identity/entity/User.kt
```diff
@@ -41,9 +39,24 @@ data class User(
...
+    @Column(name = "last_visit_at", nullable = false)
+    val lastVisitAt: LocalDateTime = LocalDateTime.now(),
+
+    @Enumerated(EnumType.STRING)
+    @Column(name="status", nullable = false)
+    val status: Status
```
Joseph-Sameh-0, line 51: note: with this new non nullable field the auto migration will not append this column
you should do the migration manually on the database with providing default value for existing row

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -79,4 +82,20 @@ class UserService(
...
+
+    fun findUsersByQuery(query: String, pageable: Pageable): Page<User> {
+        return userRepository.findByFullNameOrPhoneNumber(query, pageable)
+    }
+
+    fun updateUserLastLoginTime(userId: UUID, time: LocalDateTime){
+        val user = findById(userId)
+        val updatedUser = user.copy(lastLoginAt = time)
+        userRepository.save(updatedUser)
+    }
```
Joseph-Sameh-0, line 107: why you don't make updates query directly?

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -79,4 +82,20 @@ class UserService(
...
+
+    fun updateUserLastVisitTime(userId: UUID, time: LocalDateTime){
+        val user = findById(userId)
+        val updatedUser = user.copy(lastVisitAt = time)
+        userRepository.save(updatedUser)
+    }
```
Joseph-Sameh-0, line 113: also here 
why you don't make updates query directly?

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
MuhammedEdrees: I followed the same style used in other functions in the same service for consistency. I think they use findById just to ensure that user actually exists and if not it throws `UserNotFoundException`.

identity/src/main/kotlin/net/thechance/identity/entity/User.kt
MuhammedEdrees: Yeah, we will need to manually run alter query to add default values for these columns.

identity/src/main/kotlin/net/thechance/identity/entity/User.kt
MuhammedEdrees: We can but this will complicate the query and will require a complex join and sorting logic to be able to sort by fields in user table and other tables. I don't think this is maintainable and i'm not sure if it's performant.

Ahmed7osny1 (approved):

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
Joseph-Sameh-0: you can know if the user exist or non by knowing the affected rows number from the update function

Joseph-Sameh-0 (approved):

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
ThoriaHamdy99, line 1: Great work
 Just you need to add test cases for the new functions here

identity/src/main/kotlin/net/thechance/identity/service/AuthenticationService.kt
ThoriaHamdy99, line 1: What about modifying the test cases for this service as you modified the code

Karrar-Mohammed (approved):

ThoriaHamdy99 (approved):

Karrar-Mohammed (approved):

Karrar-Mohammed:
if we are adding the functionality to block user, then we should also add logic to the jwt fitler so with each request we check if the user is blocked or not if it blocked then we don't proceed the request and return error

Joseph-Sameh-0:
note: not just block the user actions
any action with this user will be blocked 
like transferring money to him not allowed and other actions 
so jwt alone won't satisfy this 
what do you think? 
@Karrar-Mohammed

PR #146, opened 2025-11-03
FTH-105 get mosques nearby

Mostafa-khaled21:
## what is the PR Scope ?
 fetching nearby mosques

## why do we need that ?

We need this feature to allow users to find mosques near their current location

image


faith/src/main/kotlin/net/thechance/faith/api/controller/MosqueController.kt
```diff
@@ -0,0 +1,31 @@
...
+    private val mosqueService: MosqueService
+) {
+
+    @GetMapping("/nearby")
+    fun getNearbyMosques(
+        @RequestParam latitude: Double,
+        @RequestParam longitude: Double,
+        @RequestParam(defaultValue = "10.0") radiusKm: Double
+    ): ResponseEntity<List<MosqueResponse>> {
+        if (radiusKm <= 0) throw InvalidRequestParameterException("Parameter 'radiusKm' must be greater than 0.")
```
devfalah: I don’t recommend throwing an exception here either move it to the service layer or use validation annotations instead

Mostafa-khaled21: I’ve updated the code, please take another look.
@devfalah

devfalah (approved):

marwanqashwa (approved):

mohannadahmed00 (approved):

PR #147, opened 2025-11-03
FTH-106 search mosques by name

Mostafa-khaled21:
## what is the PR Scope ?
Add search mosques endpoint

## why do we need that ?
We need this feature to enable users to search for mosques by name

image

### DIFF
--- a/faith/src/main/kotlin/net/thechance/faith/api/controller/MosqueController.kt
+++ b/faith/src/main/kotlin/net/thechance/faith/api/controller/MosqueController.kt
@@ -1,0 +1,32 @@
+package net.thechance.faith.api.controller
+
+import net.thechance.faith.api.dto.nearestMosque.MosqueResponse
+import net.thechance.faith.api.dto.nearestMosque.toMosqueResponse
+import net.thechance.faith.exception.InvalidRequestParameterException
+import net.thechance.faith.service.MosqueService
+import org.springframework.http.ResponseEntity
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("faith/mosques")
+class NearbyMosqueController(
+    private val mosqueService: MosqueService
+) {
+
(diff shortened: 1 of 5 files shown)


faith/src/main/kotlin/net/thechance/faith/api/controller/MosqueController.kt
```diff
@@ -0,0 +1,32 @@
...
+@RestController
+@RequestMapping("faith/mosques")
+class NearbyMosqueController(
+    private val mosqueService: MosqueService
+) {
+
+    @GetMapping("/search")
+    fun searchMosquesByName(
+        @RequestParam keyword: String
+    ): ResponseEntity<List<MosqueResponse>> {
```
marwanqashwa: I think this search must be paginated because we don't know how many mosques we will get what do you think?

Mostafa-khaled21: I’ve updated the code, please take another look. 
@marwanqashwa 
image

faith/src/main/kotlin/net/thechance/faith/api/controller/MosqueController.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    @GetMapping("/search")
+    fun searchMosquesByName(
+        @RequestParam keyword: String,
+        @PageableDefault(size = 10, page = 0, sort = ["name"], direction = Direction.ASC)
+        pageable: Pageable
+    ): ResponseEntity<Page<MosqueResponse>> {
+        if (keyword.isBlank()) {
+            throw InvalidRequestParameterException("Parameter 'keyword' must not be blank.")
+        }
```
devfalah: I don’t recommend throwing exception here either move it to the service layer or use validation annotations instead

marwanqashwa (approved):

faith/src/main/kotlin/net/thechance/faith/api/controller/MosqueController.kt
Mostafa-khaled21: I’ve updated the code, please take another look.
@devfalah

Hasa9nAli (approved):
Keep going Mostafa

devfalah (approved):

PR #154, opened 2025-11-04
Feature/TRN-101 video protection

mohamedd-hassan:
## what is the PR Scope ?
- we need to make links signed and expired in 1-5 minutes
- we need to make it so its only accessible from the app directly (directly pasting a valid presigned url into browser should give access denied)
- add a new endpoint for refreshing expired urls

## why do we need that ?
the first two points are for security as outlined by the user story
the third point (refresh) is so that the feature works as expected

`/trends/{trendId}/refresh`

response body
```
{
    "videoPath": "signed url for video",
    "thumbnailPath": "signed url for thumbnail"
}
```

`thumbnailPath` is nullable


app/src/main/resources/application.properties
```diff
@@ -54,4 +54,6 @@ file-size.modules.chat=5
 # === Elasticsearch  setup ===
 spring.elasticsearch.uris=${ELASTIC_SEARCH_URIS}
 spring.elasticsearch.password=${ELASTIC_SEARCH_PASSWORD}
-spring.elasticsearch.username=${ELASTIC_SEARCH_USERNAME}
\ No newline at end of file
+spring.elasticsearch.username=${ELASTIC_SEARCH_USERNAME}
+
+trends-access.secret-value=${ACCESS_SECRET}
```
ahmeddsamii: What will be the value of this in the server?

mohamedd-hassan: any value as long as its secret just like the jwt token

ahmeddsamii: Okay i think falah or karrar should put this value in the server computer,  yeah?

yusufnasserdev (approved):

trends/src/main/kotlin/net/thechance/trends/service/TrendsService.kt
```diff
@@ -44,7 +50,9 @@ class TrendsService(
                 10,
                 pageable.getSortOr(Sort.by(Sort.Direction.DESC, "createdAt"))
             )
-        ).map { it.withOwnership(currentUserId = currentUserId) }
+        ).map {
```
iZyadNasser, line 53: didn't falah tell us to not use the withOwnership function?

trends/src/main/kotlin/net/thechance/trends/service/TrendsService.kt
```diff
@@ -141,4 +151,46 @@ class TrendsService(
...
+            val trend = trendWithLikeStatus.getTrend()
+            val signedUrls = generatePresignedUrlsForTrend(
+                videoKey = trend.videoUrl, thumbnailKey = trend.thumbnailUrl
+            )
+
+            val updatedTrend = trend.copy(
+                videoUrl = signedUrls.videoUrl, thumbnailUrl = signedUrls.thumbnailUrl
+            )
+
+            return object : TrendWithLikeStatus {
```
iZyadNasser: maybe separate the making of the object to another function, instead of creating the anonymous object right after the return it's not very readable for me

iZyadNasser (approved):

app/src/main/resources/application.properties
devfalah:
```suggestion
trends-access.secret-value=${STORAGE_TRENDS_ACCESS_SECRET}
```

devfalah (approved):

trends/src/main/kotlin/net/thechance/trends/service/TrendsService.kt
mohamedd-hassan: it was to move it from the controller (not its responsibility to determine who owns what) to the service

trends/src/main/kotlin/net/thechance/trends/service/TrendsService.kt
mohamedd-hassan: the issue is trend with like status is an interface made for a feature named projections used by spring boot jpa repository so unlike a normal data class to return it we need to implement it and return the implementation

iZyadNasser: I mean just for readability, separate the creation to a private function

PR #155, opened 2025-11-05
Feature/FTH-156 Get Surah ZIP URL

marwanqashwa:
## what is the PR Scope ?
Add new endpoint to get url of zip file contain all ayat of surah
Add time out duration for remote client
## why do we need that ?
We need it in tilawah for caching all ayat local
image


Asiasama710 (approved):

Hasa9nAli (approved):

faith/src/main/kotlin/net/thechance/faith/service/tilawah/ReciterUrlGenerator.kt
```diff
@@ -0,0 +1,17 @@
...
+class ReciterUrlGenerator {
+
+    fun generateAyahSoundUrl(surahNumber: Int, ayahNumber: Int, reciter: Reciter): String {
+        return "${reciter.serverUrl}${surahNumber.padWithThreeDigits()}${ayahNumber.padWithThreeDigits()}.mp3"
+    }
+
+    fun generateSurahSoundsUrl(surahNumber: Int, reciter: Reciter): String {
+        return "${reciter.serverUrl}zips/${surahNumber.padWithThreeDigits()}.zip"
+    }
+}
```
Karrar-Mohammed: i suggest to create interface to this class

Karrar-Mohammed (approved):

PR #159, opened 2025-11-06
Feature/wlt 118 handle blocked user

MuhammedEdrees:
## what is the PR Scope ?
- Handle blocked user during login to return `FORBIDDEN` if user is blocked by admin.
- Updated status code returned when user is ip blocked to be `TOO_MANY_REQUESTS`.
- Added `UserStatusUpdatedEvent` publishing when user status changes and handled it in wallet module.
- Added validation on whether user is blocked or not during transaction initiation and completion processes.

## why do we need that ?
- To stop blocked users from using the application/performing any un-authorized operations


AhmadKharfan (approved):

Ahmed7osny1 (approved):

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -119,9 +122,16 @@ class UserService(
...
+        todo: if the user is blocked call logout function to invalidate his refresh token
          */
     }
+
+    private fun User.Status.toEventStatus(): UserStatusUpdatedEvent.UserStatus {
+        return when (this) {
+            User.Status.ACTIVE -> UserStatusUpdatedEvent.UserStatus.ACTIVE
+            User.Status.BLOCKED -> UserStatusUpdatedEvent.UserStatus.BLOCKED
+        }
+    }
```
Joseph-Sameh-0: the mapping is not the responsibility of the service

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -119,9 +122,16 @@ class UserService(
     fun updateUserStatus(userId: UUID, status: User.Status) {
         val updatedUserCount = userRepository.updateStatus(userId, status)
         if (updatedUserCount == 0) throw UserNotFoundException("User with id: $userId not found")
+        eventPublisher.publish(UserStatusUpdatedEvent(userId, status.toEventStatus()))
         /*
-        todo: should send event to notify other modules about user status change
-         and if the user is blocked call logout function to invalidate his access token
+        todo: if the user is blocked call logout function to invalidate his refresh token
          */
```
Joseph-Sameh-0: why you didn't call it instead of the comment ?

wallet/src/main/kotlin/net/thechance/wallet/service/PaymentService.kt
```diff
@@ -40,6 +40,10 @@ class PaymentService(
         if (userId == pendingTransaction.receiver.userId)
             throw IllegalArgumentException("Sender and receiver cannot be the same")
 
+        if (pendingTransaction.sender.isBlocked()) throw IllegalArgumentException("Sender is blocked")
+
+        if (pendingTransaction.receiver.isBlocked()) throw IllegalArgumentException("Receiver is blocked")
+
```
Joseph-Sameh-0: it shouldn't be invalid transaction in this case or what do you think?

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
MuhammedEdrees: It's not implemented yet. We can implement a simple logout function that just removes the refresh token.

wallet/src/main/kotlin/net/thechance/wallet/service/PaymentService.kt
MuhammedEdrees: I think a blocked user shouldn't be able to send or receive transactions. Why do you think it shouldn't be invalid?

ABDULLAHHG (approved):

wallet/src/main/kotlin/net/thechance/wallet/service/PaymentService.kt
Joseph-Sameh-0: you didn't get it right
I mean why you didn't make this transaction invalid in the database ?
or the sender can send it later ?

identity/src/main/kotlin/net/thechance/identity/service/RefreshTokenService.kt
```diff
@@ -24,4 +24,8 @@ class RefreshTokenService(
         val expiryDate = Instant.ofEpochSecond(stored.expiresIn)
         return if (expiryDate.isAfter(Instant.now())) stored else null
     }
+
+    fun deleteUserRefreshTokens(userId: UUID) {
+        refreshTokenRepository.removeByUserId(userId)
+    }
```
Joseph-Sameh-0, line 32: did you call it without `@Transactional` annotation and it didn't throw?

wallet/src/main/kotlin/net/thechance/wallet/service/PaymentService.kt
Karrar-Mohammed: why don't we just add filter or logic to the jwt filter so the blocked user will not be able to make any request, so we don't need to add check on if the user is blocked or not in every place

Karrar-Mohammed (commented):
i don't have techincal issue with this PR, but i was thinking about how to prevent the blocked user which by just add filter to the request and if the user is blocked then we stop the request and return an error?

can we do that ? what do you think ?

Joseph-Sameh-0:
it sounds good but we don't want to block the requester only 
for example: any user can't transfer money to blocked user 
and can't chat with him 
that's why we send event 
how can we achieve this cases in your opinion?
@Karrar-Mohammed

wallet/src/main/kotlin/net/thechance/wallet/service/PaymentService.kt
MuhammedEdrees: @Joseph-Sameh-0 If you mean invalid as in he won't be able to initiate a transaction and add it to pending this is already handled i just thought in some scenarios admin might block user between the transaction initiation and the payment completion so i added extra check when confirming payment.

MuhammedEdrees: @Karrar-Mohammed Hmm, that would solve the issue of blocking the user from using the system but as joseph said it won't stop other active user from interacting with the user so I think we will still need the extra check. Would this not be an issue to read from the database with each request?

MuhammedEdrees:
In addition to what @Joseph-Sameh-0 said i think it's a performance overhead if we check with each request since blocked user can't generate new access tokens it will only be an issue if he is blocked while his current access token has not expired.

identity/src/main/kotlin/net/thechance/identity/service/RefreshTokenService.kt
MuhammedEdrees: Yes, it's working fine

wallet/src/main/kotlin/net/thechance/wallet/service/PaymentService.kt
Joseph-Sameh-0: @MuhammedEdrees 
I mean mark the transaction as failed

Karrar-Mohammed: yes what i mean is to block the user from using the system but interacting with blocked user need handling but this handling will be only in some places not in every end point

Karrar-Mohammed:
> it sounds good but we don't want to block the requester only for example: any user can't transfer money to blocked user and can't chat with him that's why we send event how can we achieve this cases in your opinion? @Karrar-Mohammed

i'm not saying we will remove the event my point is about blocking the user from using the system maybe i misunderstood the usage of the event at the beginning but i think we will still need it, but if we didn't add filter or check in the jwt filter then the blocked user might still be able to use the system or some end point

identity/src/test/kotlin/net/thechance/identity/service/UserServiceTest.kt
```diff
@@ -24,7 +24,7 @@ class UserServiceTest {
     private val identityImageStorageService: IdentityImageStorageService = mockk(relaxed = true)
     private val authenticationService: AuthenticationService = mockk(relaxed = true)
     private val eventPublisher: MenaEventPublisher = mockk(relaxed = true)
-    private val userService = UserService(userRepository, identityImageStorageService, authenticationService, eventPublisher, "profile-images")
+    private val userService = UserService(userRepository, identityImageStorageService, eventPublisher, "profile-images", authenticationService)
```
Karrar-Mohammed: why don't you use the named parameter to fix this issue and it will be more readable

Karrar-Mohammed (approved):

Malak187 (approved):

Joseph-Sameh-0 (approved):

PR #162, opened 2025-11-06
Feature: IDN-152 publish events on register or update user

ThoriaHamdy99:
## what is the PR Scope ?
- Publish events when user registered and when any of the user data updated.

## why do we need that ?
- To let other features listen to these changes and update their tables.

- No end points.


poula22 (approved):

eslam-99 (approved):

AhmadKharfan (approved):

events/src/main/kotlin/net/thechance/events/identity/UserUpdatedEvent.kt
```diff
@@ -0,0 +1,33 @@
...
+    val phoneNumber: String? = null,
+    val password: String? = null,
+    val firstName: String? = null,
+    val lastName: String? = null,
+    val username: String? = null,
+    val imageUrl: String? = null,
+    val birthDate: LocalDate? = null,
+    val gender: Gender? = null,
+    val status: Status? = null,
+    val changedFields: List<ChangedField>
```
devfalah: Why do you need this?

ThoriaHamdy99: Hi Falah, I will make changes based on what we reached on our discussion at the meeting ISA

devfalah (approved):

PR #166, opened 2025-11-07
Refactor/ IDN-117 simplify rate limiter manger

poula22:
## what is the PR Scope ?
refactor rate limiter to block too many requests to the same endpoint

## why do we need that ?
to protect public apis from brute force attack


AhmadKharfan (approved):

identity/src/main/kotlin/net/thechance/identity/service/IpRateLimitManagerService.kt
```diff
@@ -1,60 +1,113 @@
...
-import java.util.concurrent.ConcurrentHashMap
-import java.util.concurrent.TimeUnit
-import java.util.concurrent.atomic.AtomicLong
 
 @Service
 class IpRateLimitManagerService(
-    private val rateLimitProperties: RateLimitProperties
+    private val rateLimitProperties: RateLimitProperties,
+    private val logService: LoginLogService,
+    private val loginLogService: LoginLogService
```
eslam-99: It should be `RequestLogService`, right?

identity/src/main/kotlin/net/thechance/identity/service/IpRateLimitManagerService.kt
```diff
@@ -1,60 +1,113 @@
...
-            return false
-        }
+    private fun isUserBlocked(
+        ipAddress: String,
+        url: String,
+        maxValidUserAttempts: Int,
+        maxWindowTimeInSeconds: Long,
+        blockTimeInSeconds: Long
+    ): Boolean {
+        val loginLogs = logService.getLoginLogsByIpAddress(
```
eslam-99: We need to change loginLogs everywhere to requestLogs

identity/src/main/kotlin/net/thechance/identity/service/IpRateLimitManagerService.kt
```diff
@@ -1,60 +1,113 @@
...
         }
+    }
 
-        val config = rateLimitProperties.endpoints[requestPath] ?: return true
+    private fun isRequestAllowed(
+        ipAddress: String,
+        url: String,
+        config: RateLimitProperties.EndpointRateLimitConfig
+    ): Boolean {
+        return isShortTermBlock(ipAddress, url, config).not() || isLongTermBlock(ipAddress, url, config).not()
```
eslam-99: As i understand, we hit DB twice, we can create a SQL statement that checks both short & long term limits at once, also to avoid retrieving all user logs to check on memory

identity/src/main/kotlin/net/thechance/identity/security/IpRateLimitFilter.kt
```diff
@@ -19,21 +19,17 @@ class IpRateLimitFilter(
...
         filterChain: FilterChain
     ) {
-        val authHeader = request.getHeader("Authorization")
         val requestUri = request.requestURI
         val clientIp = request.remoteAddr
-
         val canRateLimited = rateLimitProperties.endpoints.containsKey(requestUri)
-
-        if (authHeader == null && canRateLimited) {
+        if (canRateLimited) {
```
Karrar-Mohammed: i don't see the name "canRateLimited" is very clear and i can't understand is it mean the user reached the limit or not, so i prefer to rename it to something like "requestLimitExceeded" so it can be clear what is it mean

identity/src/main/kotlin/net/thechance/identity/service/AuthenticationService.kt
```diff
@@ -22,21 +18,20 @@ class AuthenticationService(
...
     private val jwtService: JwtService,
     private val refreshTokenService: RefreshTokenService,
-    private val loginLogService: LoginLogService,
     private val passwordEncoder: PasswordEncoder
 ) {
 
-    fun login(phoneNumber: String, password: String, ipAddress: String): AuthResponse {
-        if (isUserIpBlocked(ipAddress)) {
-            throw UserIpIsBlockedException("User with phone Number: $phoneNumber is blocked")
-        }
```
Karrar-Mohammed, line 32: we removed this from here because the request will be added to the logs and the check if the request limit exceeded now is happening in the filters of the request ?

Karrar-Mohammed (commented):
good job 👏 , just let's fix this simple issue before merging the PR

identity/src/main/kotlin/net/thechance/identity/service/IpRateLimitManagerService.kt
```diff
@@ -1,60 +1,108 @@
...
-        val config = rateLimitProperties.endpoints[requestPath] ?: return true
+    private fun isRequestAllowed(
+        ipAddress: String,
+        url: String,
+        config: RateLimitProperties.EndpointRateLimitConfig
+    ): Boolean {
+        val longTermRequestLogs = requestLogService.getRequestLogsByIpAddress(
+            ipAddress = ipAddress,
+            url = url,
+            numberOfLogs = config.longTermAttemptsLimit
```
eslam-99, line 33: I think getting logs by long term time is better than getting constatn number of values each time, do you agree?

identity/src/main/kotlin/net/thechance/identity/service/AuthenticationService.kt
poula22: Yes, everything related to blocking ip address here is moved to rate limiter manager

identity/src/main/kotlin/net/thechance/identity/service/IpRateLimitManagerService.kt
poula22: I think limit result with number of logs will always be optimal as database searches for n of records so it stops searching after it reaches this value.

identity/src/main/kotlin/net/thechance/identity/security/IpRateLimitFilter.kt
```diff
@@ -21,8 +21,8 @@ class IpRateLimitFilter(
     ) {
         val requestUri = request.requestURI
         val clientIp = request.remoteAddr
-        val canRateLimited = rateLimitProperties.endpoints.containsKey(requestUri)
-        if (canRateLimited) {
+        val requestLimitExceeded = rateLimitProperties.endpoints.containsKey(requestUri)
+        if (requestLimitExceeded) {
             if (!ipRateLimitManagerService.isRequestAllowed(clientIp, requestUri)) {
```
Karrar-Mohammed: i think we can improve this, instead of nested if we can make it one if with using &&, what do you think ?

Karrar-Mohammed (approved):

eslam-99 (approved):

PR #172, opened 2025-11-09
add update product to be out of stock 

haideraljawasim:
## what is the PR Scope ?

edit update product to accept updating the product to be out of stock 
## why do we need that ?
we need this as the requirements changed to  accept update product to be out of stock 

NO ENDPOINT

### DIFF
--- a/dukan/src/main/kotlin/net/thechance/dukan/api/dto/product/DukanProductResponse.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/api/dto/product/DukanProductResponse.kt
@@ -13,5 +13,6 @@ data class DukanProductResponse(
     val imageUrls:List<String>,
     val quantityInCart: Int,
     val createdAt: Instant,
-    val isFavorite: Boolean
+    val isFavorite: Boolean,
+    val isOutOfStock: Boolean,
 )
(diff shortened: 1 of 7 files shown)


ElsayedMagdy122 (approved):
Great Work👏

AbdelrhmanAyman20470 (approved):

YoussefmSaber (approved):

Ahmed-AlMancy (approved):
Great job

mostafa-n3ma:
can you please explain what are you planning to preform the out of stock  process.....
like for example if the outOfStock property in the product table is boolean that means we will store another table for the real products quantity ?

the quantity I think should be a number(Int) defining the real number of the available quantity of that product in the stock


so let's say there will a table for each (productID:UUID,quantity:Int) 
then there is a val isOutOfStock in the product will be changed according to some logic updating the quantity table and when the product quantity is == 0 the isOutOfStock will be updated to = true

haideraljawasim:
> can you please explain what are you planning to preform the out of stock process..... like for example if the outOfStock property in the product table is boolean that means we will store another table for the real products quantity ?
> 
> the quantity I think should be a number(Int) defining the real number of the available quantity of that product in the stock
> 
> so let's say there will a table for each (productID:UUID,quantity:Int) then there is a val isOutOfStock in the product will be changed according to some logic updating the quantity table and when the product quantity is == 0 the isOutOfStock will be updated to = true

We discussed this in yesterday meeting  we will not care about the product quantity , the owner of dukan will manually  set the product out of stock , which what mean we don't need to track the product quantity cause even the owner can't set product quantity , only can sets out of stock or not ,
then later if the product out of stack in ui we will prevent add it to cart , also in the backend

mostafa-n3ma (approved):

devfalah (approved):

PR #174, opened 2025-11-09
feature/DKN-133-delete-product

Ahmed-AlMancy:
## what is the PR Scope ?
- Implement the delete product by the owner of the dukan (Increasing the functionality) 
## why do we need that ?
- to be able to delete products

Screenshot 2025-11-09 200010


YoussefmSaber (approved):

ElsayedMagdy122 (approved):
Great Work

dukan/src/main/kotlin/net/thechance/dukan/entity/DukanProduct.kt
```diff
@@ -33,6 +33,9 @@ data class DukanProduct(
     @Column(name = "description", nullable = false, columnDefinition = "TEXT")
     val description: String,
 
+    @Column(name = "is_deleted", nullable = true)
+    val isDeleted: Boolean= false,
```
mostafa-n3ma, line 41: why are we putting a flag isDeleted here in the product entity ?

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -188,6 +188,23 @@ class DukanProductService(
...
+    fun deleteProduct(ownerId: UUID, productId: UUID) {
+        val product = dukanProductRepository
+            .findByIdAndDukanOwnerId(productId, ownerId)
+            .orElseThrow { ProductNotFoundException() }
+
+        val deletedProduct = product.copy(
+            isDeleted = true,
+        )
+
+        dukanProductRepository.save(deletedProduct)
```
mostafa-n3ma, line 201: can you. explain why we are updating the product item in the table with isDeleted flag true 
why not. just delete it ?? !

mostafa-n3ma: If we’re going with a soft-delete approach, it doesn’t seem necessary for our current needs — we don’t have audit or recovery features in place. Right now, this just adds extra complexity

dukan/src/main/kotlin/net/thechance/dukan/service/DukanProductService.kt
```diff
@@ -188,6 +188,23 @@ class DukanProductService(
...
+
+        eventPublisher.publish(
```
mostafa-n3ma, line 203: nice usage of the event publisher by the way 👍🏻

mostafa-n3ma:
@Ahmed-AlMancy 
@YoussefmSaber 
FYI
[check this article](url)

dukan/src/main/kotlin/net/thechance/dukan/entity/DukanProduct.kt
haideraljawasim: why nullable , giving it a default value false is enough no need to set it nullable

dukan/src/main/kotlin/net/thechance/dukan/repository/DukanProductRepository.kt
```diff
@@ -3,62 +3,62 @@ package net.thechance.dukan.repository
...
 import org.springframework.data.repository.query.Param
+import org.springframework.stereotype.Repository
 import java.util.*
 
 @Repository
 interface DukanProductRepository : JpaRepository<DukanProduct, UUID> {
+
     fun existsByDukanIdAndNameIgnoreCase(dukanId: UUID, name: String): Boolean
     fun existsByShelfId(shelfId: UUID): Boolean
     fun findByIdAndDukanOwnerId(id: UUID, ownerId: UUID): Optional<DukanProduct>
```
haideraljawasim, line 21: you need to modify these functions also to get only not deleted products

haideraljawasim (approved):

dukan/src/main/kotlin/net/thechance/dukan/entity/DukanProduct.kt
AbdelrhmanAyman20470: i think its better to delete row and then didnt need to check in query

haideraljawasim (approved):

AbdelrhmanAyman20470 (approved):

devfalah (approved):

PR #175, opened 2025-11-09
Refactor/dkn 143 edit product endpoint to accept the discount

ElsayedMagdy122:
## what is the PR Scope ?
refactor edit product endpoint to accept the discount

## why do we need that ?
to update discountedPrice

PUT
/dukan/product/{productId}

# Get product with base,final price and discount
image

# Edit product with base,final price 
image

# Create product with base,final price or base only
image

# Get products in cart with base,final price
image

# Get cart totalPriceBeforeDiscount,totalPriceAfterDiscount and discount
image


dukan/src/main/kotlin/net/thechance/dukan/entity/DukanProduct.kt
```diff
@@ -30,6 +30,12 @@ data class DukanProduct(
     @Column(name = "price", nullable = false)
     val price: Double,
 
+    @Column(name = "discounted_price", nullable = true)
+    val discountedPrice: Double? = null,
+
+    @Column(name = "discount", nullable = true)
```
AbdelrhmanAyman20470: why make discount and discount price?

ElsayedMagdy122: we need price after discount to display in product details and discout percntage to banners

mostafa-n3ma: I think these are extra columns — the discount percentage belongs to view-model logic on the mobile side.
On the backend, we only need to persist the actual pricing data, not display logic.
Something like basePrice and discountPrice is enough — the UI can derive and display the percentage as needed.
We shouldn’t add DB fields just to support banner visuals.

mostafa-n3ma (commented):
Please consider my comments

dukan/src/main/kotlin/net/thechance/dukan/entity/DukanProduct.kt
ElsayedMagdy122: @mostafa-n3ma 
We need to calculate the discount percentage for each product to determine which store has the highest discount. Then, we'll compare all the stores to find the top 6 or more stores with the highest discount percentages, and use this information to calculate the response.

Do I have to calculate the discount percentage for each product and then add them up to find the top 6 or more stores with the highest discount percentages every time I need this response? Do I have to start all of this from scratch and calculate the results each time?

Wouldn't this be costly for resources every time I request this response?

mostafa-n3ma: I get your point about performance, but since discounts don’t change often, we can handle this either with a query that calculates discounts directly or a small summary table updated by a job or event when needed. No need to add extra columns and duplicate data.
@ElsayedMagdy122

mostafa-n3ma (approved):

mostafa-n3ma (approved):

dukan/src/main/kotlin/net/thechance/dukan/entity/DukanProduct.kt
```diff
@@ -30,6 +30,12 @@ data class DukanProduct(
     @Column(name = "price", nullable = false)
     val price: Double,
 
+    @Column(name = "discounted_price", nullable = true)
```
TarekIdrees: I would recommend to have something similar to the mentioned json for better understanding and to prevent any calculations mistake 
1- There is a discount: 

{
  "id": "...",
  "price": {"base": 100.00, "final": 80.00, "currency": "EUR"},
  "discount": {"type": "percent", "value": 20.00}
}

2- There is no discount: 

{
  "id": "abc123",
  "name": "Wireless Keyboard",
  "price": {
    "base": 49.99,
    "final": 49.99,
    "currency": "EUR"
  },
  "discount": {
    "type": null,
    "value": 0
  }
}

Also better if we stick with Decimal for price and discount rather than Double. 

In this way it will be much easer, understandable to us and for the client what is intend to return.

YoussefmSaber: @TarekIdrees  i had a discussion with the Havana team about what is better for the prices and found out the best thing for it is data type called `BigDecimal` as it's more accurate than `Double`, but has a small draw back that it's slower when doing operations

TarekIdrees: @YoussefmSaber 

correctness matters more than microseconds.

If we take this in consideration for the trade off here. 

Also do some research about how to combine between Kotlin and JPA to apply cashing or conversation to minimise overhead

ElsayedMagdy122: @TarekIdrees  done

dukan/src/main/kotlin/net/thechance/dukan/api/dto/cart/CartItemResponse.kt
```diff
@@ -1,13 +1,14 @@
...
+import java.math.BigDecimal
 import java.util.*
 
 data class CartItemResponse(
     val productId: UUID,
     val productName: String,
     val description: String,
     val quantity: Int,
-    val price: Double,
+    val price: BigDecimal,
```
YoussefmSaber: I just noticed, we need to also add to the cart item response the price before the discount and the price after the discount

dukan/src/main/kotlin/net/thechance/dukan/api/dto/cart/CartResponse.kt
```diff
@@ -1,9 +1,10 @@
 package net.thechance.dukan.api.dto.cart
 
 import com.fasterxml.jackson.annotation.JsonProperty
+import java.math.BigDecimal
 import java.util.*
 
 data class CartResponse(
     val id: UUID,
-    val totalPrice: Double,
+    val totalPrice: BigDecimal,
```
YoussefmSaber: and here you returns the total price before the discount and after the discount,

Ahmed-AlMancy (approved):

YoussefmSaber (approved):

dukan/src/main/kotlin/net/thechance/dukan/entity/Price.kt
```diff
@@ -0,0 +1,15 @@
...
+import java.math.BigDecimal
+
+@Embeddable
+data class Price(
+    @Column(name = "base_price", nullable = false, precision = 10, scale = 2)
+    val base: BigDecimal,
+
+    @Column(name = "final_price", nullable = false, precision = 10, scale = 2)
+    val final: BigDecimal,
+)
```
haideraljawasim, line 14: I don't see the name clear `Price` what? i suggest like name it `ProductPrice` or keep it as it but place it inside the product entity if possible

dukan/src/main/kotlin/net/thechance/dukan/api/dto/product/DukanProductResponse.kt
```diff
@@ -20,8 +23,8 @@ data class DukanProductResponse(
 data class DukanProductAdminResponse(
     val id: UUID,
     val name:String,
-    val price:Double,
-    val discountedPrice: Double?,
+    val price: BigDecimal,
+    val discountedPrice: BigDecimal,
```
haideraljawasim, line 27: what was the need to convert this to BigDecimal , I mean why we needed this now

haideraljawasim (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/dto/product/DukanProductResponse.kt
ElsayedMagdy122: Financial transactions require accuracy.
https://stackoverflow.com/questions/3413448/double-vs-bigdecimal

dukan/src/main/kotlin/net/thechance/dukan/entity/Price.kt
ElsayedMagdy122: I made the name more general because it's also used on the cart as the total price for all the products on the cart.

devfalah (approved):

PR #176, opened 2025-11-09
refactor: require client generated id for sending messages

Al-Qassim:
## what is the PR Scope ?
sending messages

## why do we need that ?
we need to let the client generate the uuid for massages, so he can always sync and check if messages are sent successfully

no new endpoint, just add one parameter (messageId: UUID) to three end point (send text message endpoint, send audio message endpoint, and send image message endpoint)


raouf-kamel (approved):

dev-bilal-azzam (commented):
I think you need to remove the default uuid from the message entity also to force getting the I'd from the client.

dev-bilal-azzam (approved):

Karrar-Mohammed (approved):

PR #178, opened 2025-11-10
Feature/wlt 148 publish transaction completed successfully event

hagarabobakr:
## what is the PR Scope ?
This PR updates the PaymentService in the wallet module to publish a new TransactionCompletedEvent after every payment process whether the transaction succeeds or fails.

## why do we need that ?
Other modules (like dukan, chat, etc.) need to be notified about all completed transactions


wallet/src/main/kotlin/net/thechance/wallet/service/mapper/mapper.kt
```diff
@@ -0,0 +1,23 @@
...
+        },
+        type = when (this.type) {
+            Transaction.Type.P2P -> TransactionCompletedEvent.TransactionType.P2P
+            Transaction.Type.ONLINE_PURCHASE -> TransactionCompletedEvent.TransactionType.ONLINE_PURCHASE
+        },
+        senderId = this.sender.userId,
+        receiverId = this.receiver.userId,
+        amount = this.amount
+    )
+}
```
Joseph-Sameh-0: you can move this to `Transaction` entity file for consistency

wallet/src/main/kotlin/net/thechance/wallet/eventListener/mapper/TransactionCompletedEventMapper.kt
Joseph-Sameh-0, line 1: you can remove this file 😂

wallet/src/main/kotlin/net/thechance/wallet/service/mapper/mapper.kt
MuhammedEdrees: I don't think it a good practice to add mapper in the entity as it creates a dependency on event. Also, if we were to add more events that require mapping from transaction it's not ideal to include all these mappers in the entity file.

Malak187: You can move it back to the service package in a mapper folder 😂 @hagarabobakr

Joseph-Sameh-0: I suggest to add it in the same file because it's related to each other if the entity changed the mappers related to it will change too
it's up to you but we should choose one approach 
@MuhammedEdrees

wallet/src/main/kotlin/net/thechance/wallet/entity/Transaction.kt
MuhammedEdrees, line 1: remove the unused import and the empty line at the end

MuhammedEdrees (approved):

wallet/src/main/kotlin/net/thechance/wallet/service/mapper/mapper.kt
MuhammedEdrees: that's fine if the entity changes but what if the event changes? i think since the event depends on the entity not the other way around normally we would include the mapper there but since it's in an external module i think it's better if we put it in separate file

Joseph-Sameh-0 (approved):

wallet/src/main/kotlin/net/thechance/wallet/service/mapper/TransactionMapper.kt
```diff
@@ -0,0 +1,23 @@
...
+        createdAt = this.createdAt,
+        status = when (this.status) {
+            Transaction.Status.FAILED -> TransactionCompletedEvent.TransactionStatus.FAILED
+            Transaction.Status.SUCCESS -> TransactionCompletedEvent.TransactionStatus.SUCCESS
+            Transaction.Status.PENDING -> TransactionCompletedEvent.TransactionStatus.PENDING
+        },
+        type = when (this.type) {
+            Transaction.Type.P2P -> TransactionCompletedEvent.TransactionType.P2P
+            Transaction.Type.ONLINE_PURCHASE -> TransactionCompletedEvent.TransactionType.ONLINE_PURCHASE
+        },
```
devfalah: Its better to move them to separate methods

devfalah (approved):

wallet/src/main/kotlin/net/thechance/wallet/service/mapper/TransactionMapper.kt
hagarabobakr: Ok

PR #189, opened 2025-11-12
Feature/wlt 156 create admin deposit endpoint

hagarabobakr:
## what is the PR Scope ?
Add a new endpoint for admin wallet deposit.

## why do we need that ?
Provides admins a way to credit users’ wallets.

```HTTP
POST /wallet/admin/balance/deposit
```
### Request Body:
```JSON
{
  "phoneNumber": "01012345678",
  "amount": 100.0
}
```
### Response
```HTTP
Status: 200 OK on success
```


wallet/src/main/kotlin/net/thechance/wallet/api/controller/AdminWalletController.kt
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import java.util.*
+
+@RestController
+@RequestMapping("/wallet/admin")
+class AdminWalletController(
+    private val adminWalletService: AdminWalletService
+) {
+    @PostMapping("/balance/deposit")
```
MuhammedEdrees: No need for balance in the path since we moved ti to separate controller

wallet/src/main/kotlin/net/thechance/wallet/api/controller/AdminWalletController.kt
```diff
@@ -0,0 +1,26 @@
...
+    fun adminDeposit(
```
MuhammedEdrees: Function name should start with a verb

wallet/src/main/kotlin/net/thechance/wallet/api/controller/AdminWalletController.kt
```diff
@@ -0,0 +1,26 @@
...
+        @AuthenticationPrincipal adminId: UUID,
+        @RequestBody request: DepositRequest
+    ): ResponseEntity<Unit> {
+        val response = adminWalletService.deposit(userId = adminId, request = request)
```
MuhammedEdrees: deposit function returns unit so we can just call it without storing the return value in a response

wallet/src/main/kotlin/net/thechance/wallet/api/dto/balance/DepositRequest.kt
```diff
@@ -0,0 +1,28 @@
+package net.thechance.wallet.api.dto.balance
+
+import net.thechance.wallet.entity.Block
+import net.thechance.wallet.entity.Transaction
+import net.thechance.wallet.entity.WalletUser
+import java.math.BigDecimal
+import java.util.*
+
+data class DepositRequest(
```
MuhammedEdrees: We can rename to AdminDepositRequest to be more specific in case with add other deposit types

wallet/src/main/kotlin/net/thechance/wallet/api/dto/balance/DepositRequest.kt
```diff
@@ -0,0 +1,28 @@
...
+    return Transaction(
+        id = UUID.randomUUID(),
+        sender = sender,
+        receiver = receiver,
+        amount = BigDecimal.valueOf(this.amount),
+        block = block,
+        status = Transaction.Status.SUCCESS,
+        type = Transaction.Type.DEPOSIT
+    )
+}
```
MuhammedEdrees: I think this function is used in the service. The service is a business layer which shouldn't depend on presentation layer so we should remove it.

wallet/src/main/kotlin/net/thechance/wallet/service/AdminWalletService.kt
```diff
@@ -0,0 +1,32 @@
+package net.thechance.wallet.service
+
+import net.thechance.wallet.api.dto.balance.DepositRequest
+import net.thechance.wallet.api.dto.balance.toTransaction
+import net.thechance.wallet.exception.BlockedWalletUserException
+import net.thechance.wallet.repository.TransactionRepository
+import net.thechance.wallet.repository.WalletUserRepository
+import java.util.*
+
+class AdminWalletService(
```
MuhammedEdrees, line 11: Maybe rename it to DepositService

wallet/src/main/kotlin/net/thechance/wallet/service/AdminWalletService.kt
```diff
@@ -0,0 +1,32 @@
...
+    private val walletUserRepository: WalletUserRepository,
+    private val blockService: BlockService
+) {
+
+    fun deposit(userId: UUID, request: DepositRequest) {
+        val block = blockService.getCurrentBlock()
+        val sender = walletUserRepository.getReferenceById(userId)
+
+        val receiver = walletUserRepository.findByPhoneNumber(request.phoneNumber)
+            ?: throw IllegalArgumentException("Receiver not found")
```
MuhammedEdrees: It's better if we move findByPhoneNumber to WalletUserService and depend on it here that way we can move any common logic like throwing exception in the service itself

wallet/src/main/kotlin/net/thechance/wallet/service/AdminWalletService.kt
```diff
@@ -0,0 +1,32 @@
...
+
+class AdminWalletService(
+    private val transactionRepository: TransactionRepository,
+    private val walletUserRepository: WalletUserRepository,
+    private val blockService: BlockService
+) {
+
+    fun deposit(userId: UUID, request: DepositRequest) {
+        val block = blockService.getCurrentBlock()
+        val sender = walletUserRepository.getReferenceById(userId)
```
MuhammedEdrees: Same here we can user the function in the wallet user service

wallet/src/main/kotlin/net/thechance/wallet/service/AdminWalletService.kt
```diff
@@ -0,0 +1,32 @@
...
+import net.thechance.wallet.repository.WalletUserRepository
+import java.util.*
+
+class AdminWalletService(
+    private val transactionRepository: TransactionRepository,
+    private val walletUserRepository: WalletUserRepository,
+    private val blockService: BlockService
+) {
+
+    fun deposit(userId: UUID, request: DepositRequest) {
```
MuhammedEdrees: A function in the service shouldn't depend on a request in the api layer you can the pass the parameters explicitly

```suggestion
    fun deposit(adminId: UUID, receiverPhoneNumber: String, amount: Double) {
```

wallet/src/main/kotlin/net/thechance/wallet/service/AdminWalletService.kt
MuhammedEdrees: Also service needs @Service annotation

PR #192, opened 2025-11-12
Feature/wlt 156 create admin deposit endpoint

hagarabobakr:
## what is the PR Scope ?
Add a new endpoint for admin wallet deposit.

## why do we need that ?
Provides admins a way to credit users’ wallets.

```HTTP
POST /wallet/admin/deposit
```
### Request Body:
```JSON
{
  "phoneNumber": "01012345678",
  "amount": 100.0
}
```
### Response
```HTTP
Status: 200 OK on success
```


wallet/src/main/kotlin/net/thechance/wallet/entity/Transaction.kt
```diff
@@ -47,6 +47,7 @@ data class Transaction(
 
     enum class Type {
         P2P,
-        ONLINE_PURCHASE
+        ONLINE_PURCHASE,
+        DEPOSIT
```
Joseph-Sameh-0, line 51: why the DEPOSIT is different from P2P?

wallet/src/main/kotlin/net/thechance/wallet/service/AdminWalletService.kt
```diff
@@ -0,0 +1,34 @@
...
+
+@Service
+class AdminWalletService(
+    private val transactionRepository: TransactionRepository,
+    private val walletUserService: WalletUserService,
+    private val blockService: BlockService
+) {
+    fun deposit(userId: UUID, receiverPhoneNumber: String, amount: Double) {
+        val block = blockService.getCurrentBlock()
+        val sender = walletUserService.getUserById(userId)
```
Joseph-Sameh-0: you can use getReferenceById for the sender

wallet/src/main/kotlin/net/thechance/wallet/api/controller/AdminWalletController.kt
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.web.bind.annotation.RestController
+import java.util.*
+
+@RestController
+@RequestMapping("/wallet/admin")
+class AdminWalletController(
+    private val adminWalletService: AdminWalletService
+) {
+    @PostMapping("/deposit")
+    fun depositToUser(
```
Malak187: What about renaming each `deposit` word in the files to `depositMoney` for more clarity 

```suggestion
    fun depositMoneyToUser(
```

Joseph-Sameh-0 (approved):

wallet/src/main/kotlin/net/thechance/wallet/service/AdminWalletService.kt
```diff
@@ -0,0 +1,34 @@
...
+import java.math.BigDecimal
+import java.util.*
+
+@Service
+class AdminWalletService(
+    private val transactionRepository: TransactionRepository,
+    private val walletUserService: WalletUserService,
+    private val blockService: BlockService
+) {
+    fun deposit(userId: UUID, receiverPhoneNumber: String, amount: Double) {
```
Malak187:
```suggestion
    fun deposit(adminId: UUID, receiverPhoneNumber: String, amount: Double) {
```

wallet/src/main/kotlin/net/thechance/wallet/service/AdminWalletService.kt
Malak187: I am also thinking that since the sender represents the admin so it differes from the wallet user, we can create new table for wallet admin users like we separated them in the identity schema. What do you think?

Malak187 (approved):
Great work Hager 👏

wallet/src/main/kotlin/net/thechance/wallet/entity/Transaction.kt
hagarabobakr: I think it's good to differentiate between transactions between users and transactions sent by the admin.

MuhammedEdrees (approved):
Well done 👏👏

wallet/src/main/kotlin/net/thechance/wallet/api/dto/balance/AdminDepositRequest.kt
```diff
@@ -0,0 +1,6 @@
+package net.thechance.wallet.api.dto.balance
+
+data class AdminDepositRequest(
+    val phoneNumber : String,
+    val amount: Double,
```
Karrar-Mohammed: can we use BigDecimal here ?

Karrar-Mohammed (approved):
for current requirement i see this implementation is good enough, but i'm concern if someone could steal admin user or steal his access token then he will be able to deposit money to any user so i think it will need to add some security or validation steps like otp or verification message to make sure the admin himself that is doing the request

PR #198, opened 2025-11-14
fix: make regex accept multiple name

AbdelrhmanAyman20470:
## what is the PR Scope ?
fix  dukan shelf name 

## why do we need that ?
for dukan shelf can accept multiple name

**when regex didnt accept  multiple name**
image

**after** 
image
image

### DIFF
--- a/dukan/src/main/kotlin/net/thechance/dukan/api/dto/shelf/DukanShelfTitleRequest.kt
+++ b/dukan/src/main/kotlin/net/thechance/dukan/api/dto/shelf/DukanShelfTitleRequest.kt
@@ -7,7 +7,7 @@ data class DukanShelfTitleRequest(
     @field:NotBlank
     @field:Size(max = 50, message = "name must not exceed 50 characters")
     @field:Pattern(
-        regexp = "^[A-Za-z-]+$",
+        regexp = "^[\\p{L}0-9]+(\\s[\\p{L}0-9]+)*$",
         message = "name can only contain letters and '-'"
     )
     val title: String,


dukan/src/main/kotlin/net/thechance/dukan/api/dto/shelf/DukanShelfTitleRequest.kt
```diff
@@ -7,7 +7,7 @@ data class DukanShelfTitleRequest(
     @field:NotBlank
     @field:Size(max = 50, message = "name must not exceed 50 characters")
     @field:Pattern(
-        regexp = "^[A-Za-z-]+$",
+        regexp = "^[\\p{L}0-9]+(\\s[\\p{L}0-9]+)*$",
```
mostafa-n3ma, line 10: Have you checked this validation?

mostafa-n3ma (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/dto/shelf/DukanShelfTitleRequest.kt
AbdelrhmanAyman20470: yes
image
image
image
image

Ahmed-AlMancy (approved):

dukan/src/main/kotlin/net/thechance/dukan/api/dto/shelf/DukanShelfTitleRequest.kt
```diff
@@ -7,7 +7,7 @@ data class DukanShelfTitleRequest(
     @field:NotBlank
     @field:Size(max = 50, message = "name must not exceed 50 characters")
     @field:Pattern(
-        regexp = "^[A-Za-z-]+$",
```
Ahmed-AlMancy, line 10: Sorry abdo but I think that it was a requirement that the user can enter a space or anything else in the shelf names

AbdelrhmanAyman20470: yes , i know but previous regex accept upper and lower case and hyphen only so when enter  space return invalid

devfalah (approved):

TarekIdrees (approved):

PR #200, opened 2025-11-14
Feature/cht 150 send ayah message

omer1998:
## what is the PR Scope ?
enable user send message of ayah type in the chat

## why do we need that ?
it is new feature in our application

websocket endpoint: /app/chat.privateAyahMessage

this will handle saving ayah message and send this message to the other user

### DIFF
--- a/chat/src/main/kotlin/net/thechance/chat/service/model/ChatSummary.kt
+++ b/chat/src/main/kotlin/net/thechance/chat/service/model/ChatSummary.kt
@@ -29,6 +29,7 @@ fun Chat.toSummary(userId: UUID, chatName: String, imageUrl:String, lastMessage:
                 is MessageContent.Text -> content.text
                 is MessageContent.Image -> "Photo"
                 is MessageContent.Audio -> "Audio"
+                is MessageContent.Ayah -> "Ayah"
             }
 
             ChatSummary.Message(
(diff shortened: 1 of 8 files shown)


chat/src/main/kotlin/net/thechance/chat/service/model/ChatSummary.kt
```diff
@@ -29,6 +29,7 @@ fun Chat.toSummary(userId: UUID, chatName: String, imageUrl:String, lastMessage:
                 is MessageContent.Text -> content.text
                 is MessageContent.Image -> "Photo"
                 is MessageContent.Audio -> "Audio"
+                is MessageContent.Ayah -> "Ayah"
```
dev-bilal-azzam, line 32: As a suggestion we can return the `ayahText` instead of just `"Ayah"`

dev-bilal-azzam (approved):

chat/src/main/kotlin/net/thechance/chat/service/model/ChatSummary.kt
Karrar-Mohammed: that might be good suggestion

Karrar-Mohammed (approved):

raouf-kamel (approved):

PR #202, opened 2025-11-15
feat: integrating event bus with the trend response

ahmeddsamii:
## what is the PR Scope ?
Integrating the user event bus 

## why do we need that ?
To get the user info to send it with the response of trend

No new endpoints

### DIFF
--- a/trends/src/main/kotlin/net/thechance/trends/api/dto/trend/TrendResponse.kt
+++ b/trends/src/main/kotlin/net/thechance/trends/api/dto/trend/TrendResponse.kt
@@ -15,8 +15,8 @@ data class TrendResponse(
     val viewsCount: Int,
     val isLiked: Boolean,
     val isCurrentUserOwner: Boolean,
-    val username: String = "The Chance",
-    val profilePictureUrl: String = "",
+    val username: String,
+    val profilePictureUrl: String?
 )
 
 fun Trend.toResponse(isLiked: Boolean = false): TrendResponse {
@@ -29,7 +29,9 @@ fun Trend.toResponse(isLiked: Boolean = false): TrendResponse {
         likesCount = likesCount,
         viewsCount = viewsCount,
         isCurrentUserOwner = false,
-        isLiked = isLiked
+        isLiked = isLiked,
+        username = owner?.username ?: "Unknown",
+        profilePictureUrl = owner?.imageUrl
     )
 }
 
@@ -43,6 +45,8 @@ fun TrendWithOwnerShipAndLikeStatus.toResponse(): TrendResponse {
         likesCount = likesCount,
         viewsCount = viewsCount,
         isCurrentUserOwner = isCurrentUserOwner,
-        isLiked = isLiked
+        isLiked = isLiked,
+        username = username,
+        profilePictureUrl = profileImageUrl
     )
 }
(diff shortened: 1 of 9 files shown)


trends/src/main/kotlin/net/thechance/trends/api/dto/trend/TrendResponse.kt
```diff
@@ -29,7 +29,9 @@ fun Trend.toResponse(isLiked: Boolean = false): TrendResponse {
         likesCount = likesCount,
         viewsCount = viewsCount,
         isCurrentUserOwner = false,
-        isLiked = isLiked
+        isLiked = isLiked,
+        username = owner?.username ?: "Unknown",
```
yusufnasserdev, line 33: Should we allow for such a trend to be on users' feed? wouldn't it be better to omit it from the response given its data is missing?

trends/src/main/kotlin/net/thechance/trends/entity/Trend.kt
```diff
@@ -16,10 +16,17 @@ data class Trend(
 
     @Column(name = "owner_id", nullable = false)
     val ownerId: UUID,
+
+    @ManyToOne(fetch = FetchType.LAZY)
+    @JoinColumn(name = "owner_id", referencedColumnName = "user_id", insertable = false, updatable = false)
+    val owner: TrendUser? = null,
```
yusufnasserdev, line 22: Why would a `Trend` have an ownerId but not an owner, why is it nullable?

trends/src/main/kotlin/net/thechance/trends/models/TrendWithOwnerShipAndLikeStatus.kt
```diff
@@ -27,6 +31,8 @@ fun TrendWithLikeStatus.withOwnership(currentUserId: UUID): TrendWithOwnerShipAn
         likesCount = trend.likesCount,
         viewsCount = trend.viewsCount,
         isLiked = isLiked,
-        isCurrentUserOwner = currentUserId == trend.ownerId
+        isCurrentUserOwner = currentUserId == trend.ownerId,
+        username = owner?.username?: "Unknown" ,
```
yusufnasserdev, line 35: Same as earlier.

yusufnasserdev (commented):
Overall great work, just a few comments for clarifications, let's discuss.

trends/src/main/kotlin/net/thechance/trends/entity/Trend.kt
ahmeddsamii: Trend can have an ownerId but not an owner if the user gets deleted from the database then the ownerId column still contains the UUID, but when jpa tries to load the related TrendUser entity, it doesn't exist anymore, so owner is null.

trends/src/main/kotlin/net/thechance/trends/api/dto/trend/TrendResponse.kt
ahmeddsamii: Yes, i think it's a good idea

mohamedd-hassan: I think its because whenever we return a Trend Entity directly we aren't using it for displaying to the user but for some logic so we don't even fetch the username even if it exists

mohamedd-hassan (approved):

devfalah (approved):

mohamedd-hassan (approved):

iZyadNasser (approved):

PR #203, opened 2025-11-15
Fix: FTH 181 implement upsert functionality for ayah bookmarks

marwanqashwa:
## what is the PR Scope ?
implement upsert functionality for ayah bookmarks to have a unique bookmark for ayah to every user
just update insertion time
## why do we need that ?
because we need to stop this uncomfortable behavior
the same

### DIFF
--- a/faith/src/main/kotlin/net/thechance/faith/service/AyahBookmarkService.kt
+++ b/faith/src/main/kotlin/net/thechance/faith/service/AyahBookmarkService.kt
@@ -3,6 +3,7 @@ package net.thechance.faith.service
 import jakarta.transaction.Transactional
 import net.thechance.faith.entity.AyahBookmark
 import net.thechance.faith.exception.AyahBookmarkNotFoundException
+import net.thechance.faith.exception.FailedToCreateBookmarkException
 import net.thechance.faith.repository.AyahBookmarkRepository
 import org.springframework.data.domain.Page
 import org.springframework.data.domain.Pageable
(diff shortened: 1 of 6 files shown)


faith/src/main/kotlin/net/thechance/faith/service/AyahBookmarkService.kt
```diff
@@ -16,7 +16,15 @@ class AyahBookmarkService(
...
-        return ayahBookmarkRepository.save(ayahBookmark)
+        ayahBookmarkRepository.upsert(ayahBookmark).let {
+            if (it <= 0) throw AyahBookmarkNotFoundException("Failed to save bookmark")
+        }
+
+        return ayahBookmarkRepository.findBySurahIdAndAyahNumberAndUserId(
+            surahId = ayahBookmark.surahId,
+            ayahNumber = ayahBookmark.ayahNumber,
+            userId = ayahBookmark.userId
+        ) ?: throw AyahBookmarkNotFoundException()
```
Karrar-Mohammed: why are we doing another query to get the bookmark that we just saved ? i suggest either we just return success to the client or try to make the upsert function return the saved bookmark

marwanqashwa: I tried to make the method return an object, but any method marked with @Modifying can only return int or void. I didn’t make it return a success value only because it was already working before, and I tried to fix the issue without affecting the other side.

marwanqashwa: But what you said makes sense this is an extra query. I’ll try to make the method just void only.

Karrar-Mohammed (approved):

Hasa9nAli (approved):

Asiasama710 (approved):

PR #204, opened 2025-11-15
fix: add phone number validation when syncing contacts

Al-Qassim:
## what is the PR Scope ?
adding validation to phone numbers when syncing the contact, so we exclude invalid contacts

## why do we need that ?
so we store only valid phone numbers, with a uniform format

no endpoint added

the image below, show a set of a valid and invalid phone numbers, and how they are stored in the database

image

all the validation is handled using a google library


raouf-kamel (approved):

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
```diff
@@ -26,14 +30,34 @@ class ContactService(
...
+            .let(::checkValidContacts)
+            .distinctBy { it.phoneNumber }
+            .takeIf { it.isNotEmpty() } ?: return
 
-        if (uniqueContacts.isEmpty()) return
-        val ownerId = contactRequests.first().contactOwnerId
-        contactRepository.deleteByContactOwnerIdAndPhoneNumbers(ownerId, uniqueContacts.map { it.phoneNumber })
-        contactRepository.saveAll(uniqueContacts)
+        val ownerId = checkedUniqueContacts.first().contactOwnerId
+        contactRepository.deleteByContactOwnerIdAndPhoneNumbers(ownerId, checkedUniqueContacts.map { it.phoneNumber })
```
Karrar-Mohammed: i have two questions here why we are getting the contact owner id from the first time in the list ? can't we just take the user id of the requester ?

second question why are we deleting the contacts before saving them ?

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
```diff
@@ -26,14 +30,34 @@ class ContactService(
...
+        contactRepository.saveAll(checkedUniqueContacts)
     }
 
+    fun checkValidContacts(contacts: List<Contact>): List<Contact> {
+        if (contacts.isEmpty()) return emptyList()
+        val ownerId = contacts.first().contactOwnerId
```
Karrar-Mohammed: also the fetching the owner id from the first item is duplicated here

chat/src/main/kotlin/net/thechance/chat/service/phone_number/GooglePhoneNumberValidatorImpl.kt
```diff
@@ -0,0 +1,48 @@
...
+            nationalNumberString
+        }
+
+        return ValidatedPhoneNumber(
+            phoneNumber = e164Format,
+            countryCode = countryCode,
+            regionCode = regionCode,
+            carrierPrefixHeuristic = carrierPrefix
+        )
+    }
```
Karrar-Mohammed: this function is too long can we split it into smaller functions ?

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
Al-Qassim: first, yes, we definitely can

second, so we don't have duplicates, and we update the names of the phone numbers

chat/src/main/kotlin/net/thechance/chat/service/phone_number/GooglePhoneNumberValidatorImpl.kt
Al-Qassim: I took this function from identity, copy paste

chat/src/main/kotlin/net/thechance/chat/service/phone_number/GooglePhoneNumberValidatorImpl.kt
```diff
@@ -0,0 +1,48 @@
+package net.thechance.chat.service.phone_number
+
+import com.google.i18n.phonenumbers.NumberParseException
+import com.google.i18n.phonenumbers.PhoneNumberUtil
+import com.google.i18n.phonenumbers.Phonenumber
+import net.thechance.chat.service.exception.InvalidPhoneNumberException
+import org.springframework.stereotype.Service
+
+@Service
+class GooglePhoneNumberValidatorImpl : PhoneNumberValidator {
```
Karrar-Mohammed: try to separate the parser in a separate class

chat/src/main/kotlin/net/thechance/chat/service/phone_number/MenaCountry.kt
```diff
@@ -0,0 +1,33 @@
+package net.thechance.chat.service.phone_number
+
+enum class MenaCountry(
```
Karrar-Mohammed: try to remove this enum class and depend on the google library

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
Al-Qassim: one, done

second:
unfortunately, jpa doesn't support an annotation for upsert for non id columns, and if I use the native upsert in the query, I'll need to unpack the class of the contacts and write all the attributes in the query, witch is not scallable

so I prefer to keep it as two queries, it simple, scalable, and it will not cause a big load on  the server, the user rarely sync his contacts

chat/src/main/kotlin/net/thechance/chat/service/ContactService.kt
Al-Qassim: done

chat/src/main/kotlin/net/thechance/chat/service/phone_number/GooglePhoneNumberValidatorImpl.kt
Al-Qassim: done

chat/src/main/kotlin/net/thechance/chat/service/phone_number/GooglePhoneNumberValidatorImpl.kt
Al-Qassim: it's just a parses, I renamed it.
the validation show happen automatically.
it is like Uuid.parse(stringUuid), if stringUuid is not valid Uuid, then an exception is thrown. same here we just parse the phone number, and if its doesn't parse, just throw exception

chat/src/main/kotlin/net/thechance/chat/service/phone_number/MenaCountry.kt
Al-Qassim: done

Karrar-Mohammed (approved):

dev-bilal-azzam (approved):

PR #206, opened 2025-11-16
Feature/idn 216 implement delete account endpoint

ABDULLAHHG:
## what is the PR Scope ?
implement soft delete for user by add `is_deleted` flag to user table 
default value is false 

and update it to true when delete the user 
there is `deleted_at` column 

## why do we need that ?
for the user to delete his/her account 

no body for the request 
only authorization  bearer  token

response with message "account deleted"

### DIFF
--- a/identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
+++ b/identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
@@ -32,7 +32,7 @@ class UserService(
 ) {
 
     fun findByPhoneNumber(phoneNumber: String): User {
-        return userRepository.findByPhoneNumber(phoneNumber) ?: throw UserNotFoundException("User not found")
+        return userRepository.findByPhoneNumberAndIsDeletedFalse(phoneNumber) ?: throw UserNotFoundException("User not found")
     }
 
     fun findById(userId: UUID): User {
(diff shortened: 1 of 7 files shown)


AimanYosofiUlt (approved):

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -133,4 +134,16 @@ class UserService(
         eventPublisher.publish(UserStatusUpdatedEvent(userId, status.toEventStatus()))
         if (status == User.Status.BLOCKED) authenticationService.logout(userId)
     }
+
```
AhmadKharfan, line 137: I suggest adding @Transactional annotation to this method for better data consistency
Without it the user soft-delete and token deletion happen in separate transactions which means if logout fails the user remains deleted but can still use their refresh tokens to access the system

identity/src/main/kotlin/net/thechance/identity/service/UserService.kt
```diff
@@ -133,4 +134,16 @@ class UserService(
...
+    fun deleteUser(userId: UUID) {
+        checkUserIsNotDeleted(userId)
+        val user = findById(userId)
+        userRepository.save(user.copy(isDeleted = true, deletedAt = LocalDateTime.now()))
+        eventPublisher.publish(UserDeletedEvent(id = user.id))
```
AhmadKharfan: I think it's better to move this event publishing after the logout call

AhmadKharfan (approved):

Devmockdev (approved):

devfalah (approved):

PR #208, opened 2025-11-16
Feature/IDN-220 relocate privacy and privacy static data

eslam-99:
## what is the PR Scope ?

Relocated privacy and privacy static data into resources

## why do we need that ?

For a cleaner code

N/A


identity/src/main/kotlin/net/thechance/identity/service/PrivacyAndPolicyService.kt
```diff
@@ -2,33 +2,31 @@ package net.thechance.identity.service
...
+        PrivacyAndPolicySectionModel(title = title, content = content)
+    }
+
+    private fun getUpdatedAt(locale: Locale): Instant {
+        val updatedAtString = messageSource.getMessage("identity.privacy.policy.updatedAt", null, locale)
+        return Instant.parse(updatedAtString)
     }
 
     companion object {
-        private val updatedAt = Instant.parse("2026-11-11T00:00:00Z")
```
devfalah, line 18: you can keep this here

devfalah (approved):

ABDULLAHHG (approved):

ThoriaHamdy99 (approved):

PR #212, opened 2025-11-16
feature: add money message 

Al-Qassim:
## what is the PR Scope ?
adding a new type of messages: Money message
and adding receiver id to the chat model 

## why do we need that ?
to listen for p2p transactions and send money messages base on them
for the receiver id in the chat model, we need it so the user in the mobile can initiate a transaction to send an amount of money to the other person in the chat

no new endpoint, but this is the new response of the new message type
```
        {
            "id": "e720935a-f0cc-4423-bc48-df74c256439e",
            "senderId": "df0c603e-22a6-4b28-9f56-1176863ac59e",
            "chatId": "1b748ea5-49c4-45a2-8c39-42ecf6e90fb9",
            "content": {
                "amount": 25.0,
                "type": "MONEY"
            },
            "reactions": [],
            "sendAt": "2025-11-16T20:22:18.214630Z",
            "updatedAt": "2025-11-16T20:22:18.214630Z",
            "isRead": false,
            "isMine": true
        }
```
and this is the new response for the chat details
```
{
    "id": "1b748ea5-49c4-45a2-8c39-42ecf6e90fb9",
    "name": "abdo abdo",
    "requesterId": "df0c603e-22a6-4b28-9f56-1176863ac59e",
    "imageUrl": "d8f34378-efd2-441e-80a7-c58c0bb244f2.jpg?time=2025-11-07T05:26:53.973302057",
    "receiverId": "d8f34378-efd2-441e-80a7-c58c0bb244f2"
}
```


chat/src/main/kotlin/net/thechance/chat/eventListener/ChatEventListener.kt
```diff
@@ -25,4 +35,39 @@ class ChatEventListener(
...
+        messageRepository.save(message)
+
+        chatService.getChatUsersIds(chat.id).forEach { chatParticipantId ->
+            messagingTemplate.convertAndSendToUser(
+                chatParticipantId.toString(),
+                PRIVATE_MESSAGES,
+                message.toResponse(chatParticipantId)
+            )
+        }
+    }
```
Karrar-Mohammed, line 54: i'm thinking of the event is like the end point is just entry point and the event listener responsibility is just to listen to the event then trigger the logic that we need to do but it's not the responsibility for the event to know how to create message and send message in the socket so i prefer to move this logic to the service

Al-Qassim: done

Karrar-Mohammed (approved):

raouf-kamel (approved):

dev-bilal-azzam (approved):

PR #213, opened 2025-11-17
Feature/WLT-179 implement event listener functions on dukan events

Malak187:
## what is the PR Scope ?
- Implement event listener functions on dukan events in wallet module

## why do we need that ?
- To update the wallet dukan table whenever any of the dukan data changes


wallet/src/main/kotlin/net/thechance/wallet/eventListener/WalletEventListener.kt
```diff
@@ -30,4 +37,30 @@ class WalletEventListener(
...
+    @EventListener
+    @Async
+    fun onDukanActivationStatusUpdated(event: DukanStatusChangedEvent){
+        event.activationStatus?.let {
+            dukanService.updateDukanActivationStatus(
+                dukanId = event.dukanId,
+                activationStatus = it.toEntityActivationStatus()
+            )
+        }
+    }
```
MuhammedEdrees, line 72: Do we need two handlers for the same event?

Ahmed7osny1 (approved):
Great Work malak

MuhammedEdrees (approved):
👏🚀

dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
```diff
@@ -71,7 +68,11 @@ class DukanService(
             val updatedUser = user.copy(dukan = dukan, updatedAt = Instant.now())
             userRepository.save(updatedUser)
 
-            return dukanRepository.save(dukan)
+            val savedDukan = dukanRepository.save(dukan)
+
+            eventPublisher.publish(savedDukan.toDukanCreationEvent())
+
+            return savedDukan
```
Karrar-Mohammed, line 75: why the wallet team modify the dukan module ?

Karrar-Mohammed (approved):

dukan/src/main/kotlin/net/thechance/dukan/service/DukanService.kt
Malak187: Hello Karrar. We agreed about the event publishing with the Dukan team, also our tasks for this sprint are all about the dukan in admin panel so we modify their module and they review it when they are available. @Karrar-Mohammed

PR #216, opened 2025-11-17
Feature/TRN-122 create watch time endpoint and make it affect user categories

mohamedd-hassan:
## what is the PR Scope ?
We collect analytics (watch time) and use it to calculate how much a user prefers a category over another one so in the next PR we can recommend it more

## why do we need that ?
so we can have more precise recommendations for user

`/trends/user/watch-time` POST
response: Empty

request:
```
data class SubmitWatchTimeRequest(
    val userId: UUID,
    @field:NotEmpty val watchTimes: List<WatchTimeDto>,
)

data class WatchTimeDto(
    val trendId: UUID,
    val watchStartTimeStamp: LocalDateTime,
    val watchEndTimeStamp: LocalDateTime,
    val videoDuration: Long,
    @field:Min(0) val percentWatched: Double,
)
```

```
{
  "userId": "3b4149af-57af-45d5-b1de-1e5b09a2debd",
  "watchTimes": [
    {
      "trendId": "a5cc8f76-355a-4739-8e63-462621e3544a",
      "watchStartTimeStamp": "2025-11-17T20:15:30",
      "watchEndTimeStamp": "2025-11-17T20:18:50",
      "videoDuration": 180,
      "percentWatched": 75.5
    }
  ]
}
```


app/src/main/kotlin/net/thechance/config/storage/StorageConfig.kt
```diff
@@ -84,6 +84,8 @@ class StorageConfig(
     fun pathStyleStorageConfiguration(): S3Configuration {
         return S3Configuration.builder()
             .pathStyleAccessEnabled(true)
+            .chunkedEncodingEnabled(false)
```
ahmeddsamii: This shouldn't be in your commits, this was just for testing to our free bucket

mohamedd-hassan: oh i accidentally commited this instead of build gradle

trends/src/main/kotlin/net/thechance/trends/entity/UserCategories.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    @Id
+    @Column(name = "category_id", nullable = false)
+    val categoryId: UUID,
+
+    @Column(name = "is_selected", nullable = false)
+    val isSelected: Boolean,
+
+    @Column(name = "affinity", nullable = false, )
+    val affinity: Int = 0,
```
yusufnasserdev, line 36: Why couldn't this be `userInterestRating` or something more descriptive? this word is not that common.

trends/src/main/kotlin/net/thechance/trends/entity/UserCategories.kt
```diff
@@ -0,0 +1,45 @@
...
+    @Id
+    @Column(name = "user_id", nullable = false)
+    val userId: UUID,
+
+    @Id
+    @Column(name = "category_id", nullable = false)
+    val categoryId: UUID,
+
+    @Column(name = "is_selected", nullable = false)
+    val isSelected: Boolean,
```
yusufnasserdev, line 33: Why would an entry exist for the category in this table if the user isn't selecting it?

trends/src/main/kotlin/net/thechance/trends/service/TrendUserService.kt
```diff
@@ -43,37 +40,167 @@ class TrendUserService(
...
 
-        val trendUser = getUserOrThrow(userId)
-        val currentCategoryIds = trendUser.categories.mapTo(mutableSetOf()) { it.id }
+        val currentCategories = userCategoryRepository.findUserCategoriesByUserId(userId)
+        val currentCategoryIds = currentCategories.map { it.categoryId }
 
-        val actualRemoved = categoriesToRemove.filterTo(mutableSetOf()) { it in currentCategoryIds }
-        val actualAdded = categoriesToAdd.filterTo(mutableSetOf()) { it !in currentCategoryIds }
+        val actualRemoved = categoriesToRemove.filterTo(mutableListOf()) { it in currentCategoryIds }
+        val changedToSelected = categoriesToAdd.filterTo(mutableListOf()) { it in currentCategoryIds }
```
yusufnasserdev: If it's already in the current categories, how is it changed to selected?

What does this list really do?

yusufnasserdev (commented):
I noticed `affinity` is heavily used throughout the PR, it's fine if you wanna keep it hence the word is in of itself describe the meaning, but it will always take something a moment to search and understand what it means.

Good work overall!

trends/src/main/kotlin/net/thechance/trends/entity/UserCategories.kt
mohamedd-hassan: previously if a user deselected a category it would be removed from the table but since I added affinity if a user accidentally deselects and reselects the category I don't want to reset their affinity calculation so it acts like a soft delete

trends/src/main/kotlin/net/thechance/trends/service/TrendUserService.kt
mohamedd-hassan: basically it goes like:
user selects the category for first time ever -> means it does not exist in table -> add new entry to user_categories
user deselects category -> means it exists in table ->  update it `is_selected` to false to not reset their affinity
user reselects a category -> means it exists in table but not selected -> update `is_selected` back to true so they wouldn't have the affinity reset

trends/src/main/kotlin/net/thechance/trends/entity/UserCategories.kt
mohamedd-hassan: i felt like rating would be misleading as the user is not actively ranking them and other words seemed to long, I came across this word while researching and it fit the description in my head while being short so it was easy to write in variables and functions etc. i might try to think of a new word though that is a bit more readable

trends/src/main/kotlin/net/thechance/trends/service/TrendUserService.kt
yusufnasserdev: It might be helpful to change the variable names then to relay this.

Something like this:

```kotlin

val deSelectedCategories
val reSelectedCategories
val newSelectedCategories

```

or

```kotlin

val existingCategoriesToDeselect 
val existingCategoriesToReselect
val newCategoriesToAdd 

```

devfalah (approved):

ahmeddsamii (approved):

yusufnasserdev (approved):

PR #223, opened 2025-11-18
Feature/CHT-167-implement weather end point

Karrar-Mohammed:
## what is the PR Scope ?
implement end point to get weather details

## why do we need that ?
to display the current weather information in home screen

{baseUrl}?weather/current?lat={lat}&lng={lng}

response:

{
    "currentTemperature": 18.5,
    "minTemperature": 17.2,
    "maxTemperature": 20.1,
    "weatherCode": 2
}


chat/src/main/kotlin/net/thechance/chat/repository/weather/WeatherRepository.kt
```diff
@@ -0,0 +1,7 @@
+package net.thechance.chat.repository.weather
+
+import net.thechance.chat.api.dto.WeatherResponse
+
+interface WeatherRepository {
+    fun getCurrentWeather(latitude: Double, longitude: Double): WeatherResponse
+}
```
omer1998, line 7: as i understand repository should return entity class am i right

chat/src/main/kotlin/net/thechance/chat/repository/weather/WeatherRepositoryImp.kt
```diff
@@ -0,0 +1,51 @@
...
+            return response.body ?: throw FetchWeatherException("Empty weather response")
+
+        } catch (e: RestClientResponseException) {
+            throw FetchWeatherException("Weather API error: ${e.statusCode}", e)
+        }
+    }
+}
+
+private fun WeatherDTO.toResponse(): WeatherResponse = WeatherResponse(
+    currentTemperature = current?.temperature ?: 0.0,
```
omer1998: i think the place of this mapper here is not appropriate

omer1998 (approved):
great job karrar

chat/src/main/kotlin/net/thechance/chat/repository/weather/WeatherRepository.kt
Karrar-Mohammed: i didn't want to create entity i thought it would be redundant but i think i will do it for the consistency

chat/src/main/kotlin/net/thechance/chat/repository/weather/WeatherRepositoryImp.kt
Karrar-Mohammed: after i will add the entity this will be mapper to the entity so it's place will be correct

chat/src/main/kotlin/net/thechance/chat/repository/weather/WeatherRepositoryImp.kt
```diff
@@ -0,0 +1,44 @@
...
+
+            if (!response.statusCode.is2xxSuccessful) {
+                throw FetchWeatherException("Weather API error: ${response.statusCode}")
+            }
+
+            return response.body ?: throw FetchWeatherException("Empty weather response")
+
+        } catch (e: RestClientResponseException) {
+            throw FetchWeatherException("Weather API error: ${e.statusCode}", e)
+        }
```
devfalah, line 42: I think its better to move this code to a separate class such as `WeatherApi`

devfalah (approved):

chat/src/main/kotlin/net/thechance/chat/repository/weather/WeatherRepositoryImp.kt
Karrar-Mohammed: i don't see it's needed because this class will be just a bridge, so for simplicity i kept this code here

raouf-kamel (approved):

chat/src/main/kotlin/net/thechance/chat/repository/weather/WeatherRepositoryImp.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import net.thechance.chat.service.exception.FetchWeatherException
+import net.thechance.chat.service.model.WeatherModel
+import org.springframework.boot.web.client.RestTemplateBuilder
+import org.springframework.stereotype.Repository
+import org.springframework.web.client.RestClientResponseException
+import org.springframework.web.client.getForEntity
+
+@Repository
+class WeatherRepositoryImp(
```
Al-Qassim, line 11: we use `Impl` not `Imp`

Al-Qassim (approved):

PR #225, opened 2025-11-20
Feature/wlt 190 update wallet user

Malak187:
## what is the PR Scope ?
- Update wallet user with dukan_id when new dukan is created.
- Update user image mappers to include base image url.

### DIFF
--- a/wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
+++ b/wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
@@ -9,6 +9,7 @@ import net.thechance.wallet.entity.Transaction
 import net.thechance.wallet.service.TransactionService
 import net.thechance.wallet.service.model.input.TransactionFilterParams
 import net.thechance.wallet.service.model.input.UserTransactionType
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.data.domain.PageRequest
 import org.springframework.data.domain.Pageable
 import org.springframework.data.domain.Sort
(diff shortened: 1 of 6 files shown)


MuhammedEdrees (approved):

MennaSrour (approved):

wallet/src/main/kotlin/net/thechance/wallet/api/dto/transaction/TransactionResponse.kt
```diff
@@ -84,15 +86,16 @@ private fun getUserType(
...
+    imageBaseUrl: String
 ): TransactionPartyInfo {
     val isDeposit = type == Transaction.Type.DEPOSIT
     val name = if (isDeposit) "MENA" else sender.userName
     val imageUrl = if (isDeposit) null else sender.imageUrl
 
     return TransactionPartyInfo(
         name = name,
-        imageUrl = imageUrl
+        imageUrl = imageUrl?.let { "$imageBaseUrl/$it" }
```
Karrar-Mohammed: i see this line is duplicated in many places

wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
```diff
@@ -23,9 +24,13 @@ import java.util.*
 @RestController
 @RequestMapping("/wallet/transactions")
 class TransactionController(
+    @Value("\${storage.mena.cdn-endpoint}") cdnEndpoint: String,
+    @Value("\${identity.resources.profile-image-directory}") profileImageDirectory: String,
     private val transactionService: TransactionService,
     private val statementPdfWriter: StatementPdfWriter,
 ) {
+    private val imageBaseUrl: String = "$cdnEndpoint$profileImageDirectory"
+
```
Karrar-Mohammed: also i don't see this the responsibility of the controller, so i suggest to create a class for generating image url, you can pass the image url that we get from the database then this class will build the base url and append the image url so we can achieve reusable logic and it's one place without any duplication also we achieve the single responsibility this class will be only responsible for generating image url, now i see this logic is duplicated and separated in different places

wallet/src/main/kotlin/net/thechance/wallet/api/controller/TransactionController.kt
```diff
@@ -49,7 +46,7 @@ class TransactionController(
                 Sort.by(Sort.Direction.DESC, Transaction::createdAt.name)
             ),
             currentUserId = userId
-        ).toResponsePage(currentUserId = userId, imageBaseUrl = imageBaseUrl)
+        ).toResponsePage(currentUserId = userId, userImageUrlBuilder = userImageUrlBuilder)
```
Karrar-Mohammed: i suggest to pass the image url here

Karrar-Mohammed (approved):

PR #230, opened 2025-11-21
Feature: FTH-196 Implement end point to get list of Ayat URLs for a specific surah

Pixelh97:
## what is the PR Scope ?
Adding a new end point in the Tilawah to get a list of Ayat for a specific Surah

## why do we need that ?
To play a whole Surah from the client side.

image


Hasa9nAli (approved):

faith/src/main/kotlin/net/thechance/faith/utils/getSurahAyatNumber.kt
```diff
@@ -0,0 +1,121 @@
+package net.thechance.faith.utils
+
+fun getSurahAyatNumber(surahNumber: Int): Int {
```
marwanqashwa, line 3: we can make one test case to check the hash of this file or any way, because if one number is changed we can't know that easily

marwanqashwa (approved):

Mostafa-khaled21 (approved):

devfalah (approved):

PR #243, opened 2025-11-27
Fix/wlt 203 timezone handling inconsistencies

MuhammedEdrees:
## what is the PR Scope ?
- Improve timezone handling across all endpoints that deal with dates in wallet feature.

## why do we need that ?
- Statement endpoint was returning timestamps in server time not client time and first date endpoint were not returning it accounting for client timezone as well.

#### Before
```bash
GET wallet/transactions/statement?type=DEPOSIT&from=2024-01-01T00:00:00&to=2024-01-31T23:59:59
```

#### After
```bash
GET wallet/transactions/statement?type=DEPOSIT&from=2024-01-01&to=2024-01-31
Accept-Timezone: Africa/Cairo
```

**Changes:**
- Added `Accept-Timezone` header (defaults to UTC)
- Changed from `LocalDateTime` to `LocalDate` (no time component needed)
- Parameters: `from` and `to` now use date format `YYYY-MM-DD` instead of `YYYY-MM-DDTHH:mm:ss`

---

#### Before
```bash
GET wallet/transactions/first-date
```

#### After
```bash
GET wallet/transactions/first-date
Accept-Timezone: Asia/Baghdad
```

**Changes:**
- Added `Accept-Timezone` header (defaults to UTC)
- Response will now respect the client's timezone

---

#### Before
```bash
GET wallet/transactions?type=WITHDRAWAL&status=COMPLETED&from=2024-01-01T00:00:00&to=2024-01-31T23:59:59&page=0&size=20
```

#### After
```bash
GET wallet/transactions?type=WITHDRAWAL&status=COMPLETED&from=2024-01-01&to=2024-01-31&page=0&size=20
Accept-Timezone: Asia/Tokyo
```

**Changes:**
- Added `Accept-Timezone` header (defaults to UTC)
- Changed from `LocalDateTime` to `LocalDate` for `from` and `to` parameters
- Date parameters simplified to `YYYY-MM-DD` format

### DIFF
--- a/wallet/src/main/kotlin/net/thechance/wallet/api/controller/util/StatementHtmlGenerator.kt
+++ b/wallet/src/main/kotlin/net/thechance/wallet/api/controller/util/StatementHtmlGenerator.kt
@@ -2,6 +2,7 @@ package net.thechance.wallet.api.controller.util
 
 import net.thechance.wallet.entity.Transaction
 import net.thechance.wallet.service.model.output.StatementData
+import net.thechance.wallet.service.utils.toClientZone
 import org.springframework.data.domain.Page
 import org.springframework.stereotype.Component
 import org.thymeleaf.TemplateEngine
@@ -9,6 +10,7 @@ import org.thymeleaf.context.Context
 import java.math.BigDecimal
 import java.text.DecimalFormat
 import java.time.LocalDateTime
+import java.time.ZoneId
 import java.time.format.DateTimeFormatter
 import java.util.*
 
@@ -19,10 +21,11 @@ class StatementHtmlGenerator(
 
     fun generateForPage(
         statementData: StatementData,
+        timezone: ZoneId,
         transactionsPage: Page<Transaction>,
     ): String {
         val formattedTransactions = transactionsPage.content.map {
-            formatTransaction(it, statementData.userId)
+            formatTransaction(it, statementData.userId, timezone)
         }
 
         val templateData = mapOf(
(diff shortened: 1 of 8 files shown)


EL-MANCY (approved):

wallet/src/main/kotlin/net/thechance/wallet/api/controller/util/StatementHtmlGenerator.kt
```diff
@@ -98,9 +101,12 @@ class StatementHtmlGenerator(
...
 
-    private fun LocalDateTime.formatRowItemDate(): String =
-        this.format(DateTimeFormatter.ofPattern("yyyy/MM/dd"))
+    private fun LocalDateTime.formatRowItemDate(timezone: ZoneId): String =
+        this.toClientZone(timezone)
+            .format(DateTimeFormatter.ofPattern("yyyy/MM/dd"))
+
+    private fun LocalDateTime.formatRowItemTime(timezone: ZoneId): String =
+        this.toClientZone(timezone)
+            .format(DateTimeFormatter.ofPattern("hh:mm a"))
```
Joseph-Sameh-0, line 110: I think it's better to separate this to separate until file

wallet/src/main/kotlin/net/thechance/wallet/api/controller/util/StatementPageEventHandler.kt
```diff
@@ -140,5 +143,5 @@ class StatementPageEventHandler(
     }
 
     private fun LocalDateTime.formatHeaderDate(): String =
-        this.format(DateTimeFormatter.ofPattern("MMM dd yyyy"))
+        this.toClientZone(timezone).format(DateTimeFormatter.ofPattern("MMM dd yyyy"))
```
Joseph-Sameh-0, line 146: same

wallet/src/main/kotlin/net/thechance/wallet/api/controller/util/StatementHtmlGenerator.kt
MuhammedEdrees: I think it's fine to keep locally to the util file it's used in since there is not that much logic and it's a simple formatting function.

Joseph-Sameh-0 (approved):

Malak187 (approved):

devfalah (approved):

PR #246, opened 2025-11-28
Migrate Reciters’ Audio ZIPs to S3 on First Deployment

Mostafa-khaled21:
## what is the PR Scope?

This PR introduces a one-time migration script that automatically uploads all reciters’ audio ZIP files to our S3 bucket upon the first deployment

Notes:

This is a one-time operation and does not re-run on subsequent deployments unless the marker file is removed.

### DIFF
--- a/faith/src/main/kotlin/net/thechance/faith/service/tilawah/ReciterAudioToS3MigrationRunner.kt
+++ b/faith/src/main/kotlin/net/thechance/faith/service/tilawah/ReciterAudioToS3MigrationRunner.kt
@@ -1,0 +1,78 @@
+package net.thechance.faith.service.tilawah
+
+import net.thechance.faith.repository.RecitersRepository
+import net.thechance.faith.service.config.FaithStorageProperties
+import org.springframework.boot.ApplicationArguments
+import org.springframework.boot.ApplicationRunner
+import org.springframework.stereotype.Component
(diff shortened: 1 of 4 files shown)


Hasa9nAli (approved):

faith/src/main/kotlin/net/thechance/faith/service/tilawah/ReciterAudioToS3MigrationRunner.kt
```diff
@@ -0,0 +1,78 @@
...
+import java.time.LocalDateTime
+
+@Component
+class ReciterAudioToS3MigrationRunner(
+    private val recitersRepo: RecitersRepository,
+    private val props: FaithStorageProperties,
+    private val menaS3Client: S3Client
+) : ApplicationRunner {
+
+    private val markerFile = File("migration_done.marker")
```
Karrar-Mohammed, line 23: currently we are checking if the migration done or not by creating this file so again if the server run again and this file is exist then that mean the migration is already done before i think this is the idea,

so if i understood it correctly, then there is an issue as far as i know with each deployment all the files from the old deployment will be deleted from the server so with every new deployment this file will not be exist and this code will always run with every new deployment.

usually i think these kind of scripts it's run outside the backend we don't need to merge this code into the backend we can just build it then run it once that my suggestion about how we can do this

